PR #2, opened 2019-05-01
Jump dev

sakshi-mishra:
Hi all, 
- dat_file_manager.py is changed to create two different dictionaries for bau and with_tech data. Also, the double nesting of the dictionaries (with original datfile names) has been removed as it didn't seem relevant. 
- Currently, the wtch_data and bau_data dictionaries are getting dumped into JSON. 
- JSON being read back in the Julia code
- Nick D has tested several ways of calling Julia. We are leaning toward pyjulia package (can discuss more in person). 
- Eventually, we will incorporate the celery task management along with pyjulia to send these dictionaries to Julia code I suppose (?). 
- The structs in reo_struct.jl have been tested by creating test objects 

@nickdiorio wondering if we should merge it now or keep working on this branch (jump_dev) until the JuMP code is complete and then merge it to develop. Thoughts?

thanks!

### DIFF
--- a/reo_jump/reo_struct_test.jl
+++ b/reo_jump/reo_struct_test.jl
@@ -1,0 +1,48 @@
+#=
+reo_struct_test:
+- Julia version: 1.0.3
+- Author: Sakshi Mishra
+- Date: 2019-04-24
+=#
+
+using Revise
+using Dates
+include("utils.jl")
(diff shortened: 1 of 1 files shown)


NLaws:
@sakshi-mishra do we want to keep all of these old PR's?

PR #30, opened 2020-03-03
Speed up python -> julia interface

jmpohl:
Looking to implement `pyfunction()` or `pyfunctionret()` detailed in [this section of the PyCall documentation](https://github.com/JuliaPy/PyCall.jl#calling-julia-from-python). Don't know the full repercussions, but I think it is the next thing to try to get a speed up.


jmpohl:
Initial tests aren't the most promising. There aren't really noticeable speed-ups, but I want to make sure there isn't anything else here. Regardless, it will give us greater control over the julia type conversion, but it does just add more code that needs to be maintained.

Still need to:

- [ ] Fix tests failing because of type conversion
- [ ] Figure out PyObject conversion of the Julia Model that gets passed in 
- [ ] Do more robust testing of the whole stack

sakshi-mishra:
@jmpohl - is this branch still under development? Or is it ready for review?

jmpohl:
@sakshi-mishra still under development, and hopefully will be picked back up in a couple weeks.

PR #44, opened 2020-04-14
Multiple pv

TK-23:
## Changes

This PR allows REopt to solve for multiple PV systems by including a list of PV dictionaries instead of a single dictionary.

New PV inputs:
  - pv_name [optional] user friendly pv description
  - pv_number [internal use only] used to tie optimization results to database records
  - location [roof, ground, both] used in constraining PV sizes to area limits

Also, validator by default returns a list of PV's for internal processing. In delivering results, if there is only one PV, then it is returned as a dictionary not a list.


NLaws:
@TK-23 is this PR ready for review?

TK-23:
@nlaws yes, ready for review

TK-23:
@NLaws all set here

PR #53, opened 2020-05-22
clean up utils.jl

NLaws:
- clarifying data types in the Parameter struct by removing extraneous methods
- removing Parameter field types for AxisArrays until we finalize data types
- Still need to address many empty arrays for BAU scenarios


zolanaj:
This looks great, thanks for the edits.
- A BAU test that includes a generator and PV yields the same obj value as the reformulation branch after merging one bug fix from reformulation, so the cleanup hasn't changed anything in that setting
- I'm currently working on adding !isempty() checks prior to any sum, minimum, or maximum operators on arrays for the BAU cases, which is currently causing the JuMP model to fail
- Generally speaking, the new build_param that includes only the dictionary instead of ordered args will make the eventual CHP and TES integration a lot easier. Thanks Nick!

PR #56, opened 2020-05-30
Migrate julia pararms

zolanaj:
- [x] Tests for the changes have been added (for bug fixes / features)  (NA)
- [x] Docs have been added / updated (for bug fixes / features)  (NA)

Migrating Julia-generated parameters to the API for generation and the Julia model for loading.

- Same as the reformulation branch in terms of tests passing.

- Some more sets (NMILRegime) are now empty when the collection of technologies is also empty, such as in BAU cases.  

None that aren't already in reformulation,

* **Other information**:


reo/src/utils.jl
```diff
@@ -235,13 +235,13 @@ Base.@kwdef struct Parameter
      GridExportRates
      FuelBurnSlope
      FuelBurnYInt
-     MaxGridSales::Array{<:Real, 1}
```
NLaws, line 238: @zolanaj can we keep this type for MaxGridSales?

zolanaj: I thought it could be a good feature to impose different limits between net-metering and wholesale sales.  This would actually be different from the existing formulation, which treats all "1W" destinations the same.

For now, let's keep MaxGridSales under this type, and we can adjust the constraint to fit the original formulation.  I'll push a commit to that end shortly.

PR #59, opened 2020-06-03
Enable existing diesel gen for financial outage sim

sakshi-mishra:
@NLaws - this one enables the modeling of existing diesel generator for the financial case's outage simulation.

### DIFF
--- a/reo/src/outage_costs.py
+++ b/reo/src/outage_costs.py
@@ -58,12 +58,47 @@ def calc_avoided_outage_costs(data, present_worth_factor, run_uuid):
     site_outputs = data['outputs']['Scenario']['Site']
     pvs = site_outputs['PV']
     wind = site_outputs['Wind']
-    generator = site_outputs['Generator']
+
     load_profile = site_inputs['LoadProfile']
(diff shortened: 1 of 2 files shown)


reo/src/outage_costs.py
```diff
@@ -58,12 +58,47 @@ def calc_avoided_outage_costs(data, present_worth_factor, run_uuid):
...
+    generator_in = site_inputs['Generator']
+    generator_out = site_outputs['Generator']
+
+  # Generator is not allowed to run during normal time
+
+    if generator_in["generator_only_runs_during_grid_outage"]:
+
+        if load_profile.get('outage_start_hour') is None and load_profile.get('outage_end_hour') is None:
+            # handles financial scenario where there will only be non-zero input kw
+            diesel_kw_for_case = generator_in.get('existing_kw')
```
NLaws, line 90: @sakshi-mishra should this `.get` and the two below it have a default value of zero? (Or does some later logic require that `diesel_kw_for_case` can be `None`?)

sakshi-mishra: @NLaws - yes, the default value for existing_kw is set to zero in the nested_inputs.py.

PR #67, opened 2020-06-25
reopt.jl refactor

NLaws:
split the ~900 line `reopt_run` function into many smaller functions in order to decrease model build time


NLaws:
@zolanaj and @Bill-Becker this PR moves the constraints into functions to make the model build more efficient. Could be even more efficient if we put each constraint in its own function but this is a good start. It did reduce the average run times by 3-5 seconds on average.  These changes will make it difficult to merge the CHP branch.

PR #80, opened 2020-07-22
New outputs: emissions, IRR, LCOE, SPP; new storage rebates per kWh

TK-23:
This incorporates Private PR's 330 and 328

----- Private PR 330 ---------
## New

   ElectricTariff emissions factor series input (defaults to AVERT regional data)
   Generator emissions factor (defaults to diesel emissions factor)
   ElectricTariff, Generator and Site year 1 emissions totals as outputs
   Emissions metrics to proforma

----- Private PR 328 ---------
## New

Added to the proforma spreadsheet:
 - PV LCOE for new capacity
 - Wind LCOE 
  With  supporting calculations:
     -- Host Present Worth Factor
     -- Developer Present Worth Factor
     -- PV Energy Levelization Factor
     -- Wind Energy Levelization Factor
- Simple Payback (Developer Payback in 2 party case) in summary on main page

Added to the API outputs:
  - PV and Wind LCOE (following spreadsheet logic)
  - Simple Payback (following spreadsheet logic)
  - IRR (following spreadsheet logic)

Added to the API inputs:
 - total storage rebates per kWh

## Bug Fixes
Updated storage to read per kW and per kWh incentives from the API response
Updated final cashflow to include non-taxed year 0 incentives (CBI and IBI)


NLaws:
@TK-23 do we need proforma/REoptCashFlowTemplate.xlsm ?

reo/nested_inputs.py
```diff
@@ -1093,7 +1102,11 @@ def list_of_str(input):
           "max": 1.0e9,
           "default": 0.0,
           "description": "Maximum system size for which production-based incentives apply"
-        }
+        },
+         "emissions_factor_lb_CO2_per_gal": {
+          "type": "float",
+          "description": "Average carbon dioxide emissions factor"
```
NLaws, line 1108: @TK-23 what is the average taken over? eg. Average annual, average hourly, or average over time step (could be 15 or 30 minutes)? Please update the description.

reo/nested_outputs.py
```diff
@@ -164,7 +174,21 @@ def list_of_string(input):
                   "type": float,
                   "description": "Net replacement costs for all technologies, in future value, excluding incentives.",
                   "units": "$"
-                }
+                },
+                "simple_payback_years": {
+                  "type": float,
+                  "description": ("Number of years until the cumulative cashflow is positive. If there are charges "
```
NLaws, line 180: I had to read this 3 or 4 times to get the meaning. Can you please re-word it to make it more clear? Maybe something like "Number of years until the cumulative annual cashflow turns positive. If the cashflow becomes negative again after becoming positive then simple payback is increased by the number of years that the cash flow is negative beyond the break-even year."  And now that I've written that, could it be defined as "Number of years that the cumulative annual cashflow is negative, including negative years beyond the break-even year if they exist (for example when paying for storage replacement costs)."

reo/nested_outputs.py
```diff
@@ -455,6 +489,16 @@ def list_of_string(input):
                   "type": float,
                   "description": "Year one energy supplied from grid to load in the business-as-usual scenario",
                   "units": "kWh"
+                },
+                "year_one_emissions_lb_C02": {
+                  "type": int,
+                  "description": "Total equivalent pounds of carbon dioxide emitted from utility electricity use in the first year. Calculated by default from hourly emissions estimates except in AK and HI.",
```
NLaws, line 495: what do we calculate in AK and HI?

reo/process_results.py
```diff
@@ -70,6 +74,280 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
...
+                depreciation_schedule[0] += (storage['macrs_bonus_pct'] * macrs_bonus_basis)
+                total_depreciation += depreciation_schedule
+
+        #calculate Generator capital costs, o+m costs, incentives, and depreciation
+        if (generator['size_kw'] or 0) > 0:
+            total_kw = generator.get('size_kw') or 0
+            new_kw = (generator.get('size_kw') or 0) - (generator.get('existing_kw') or 0)
+            existing_kw = generator.get('existing_kw') or 0
+            
+            capital_costs = (new_kw * generator['installed_cost_us_dollars_per_kw'])
```
NLaws, line 251: the generator capital_costs are never used. where are any of the tech capital costs accounted for in the cash flow?

reo/process_results.py
```diff
@@ -234,6 +510,72 @@ def upfront_capex_after_incentives(self):
...
+                discount_pct = financials.offtaker_discount_pct
+
+            new_kw = (tech_results_dict.get('size_kw') or 0) - (tech_inputs_dict.get('existing_kw') or 0) # new capacity
+            
+            if new_kw == 0:
+                return None
+
+            capital_costs = new_kw * tech_inputs_dict['installed_cost_us_dollars_per_kw'] # pre-incentive capital costs
+            
+            annual_om = new_kw * tech_inputs_dict['om_cost_us_dollars_per_kw'] # NPV of O&M charges escalated over financial life
```
NLaws, line 536: should we not include tax benefits in the LCOE calculation? For example the O&M costs will be less by the tax rate if the site pays taxes.

reo/process_results.py
```diff
@@ -234,6 +510,72 @@ def upfront_capex_after_incentives(self):
...
+            output
+            :param tech_results_dict: dict of model results (i.e. outputs from PVModel )
+            :param tech_inputs_dict: dict of model results (i.e. inputs to PVModel )
+            :param financials: financial model storing input financial parameters
+            :return: float, LCOE in US dollars per kWh
+            """
+            years = financials.analysis_years # length of financial life
+
+            if financials.two_party_ownership:
+                discount_pct = financials.owner_discount_pct
```
NLaws, line 525: You probably already sorted this with Dan, but just want to make sure: I believe that what is being calculated in the two party case here is the LCOE for the developer. In order to account for the developer's required rate of return (i.e. discount rate) I think that we should still discount items with the offtaker's rate but then multiply by the two party factor, which would calculate the offtaker's LCOE with the developer's "mark-up". Definitely a Dan question though (and `@dolis` doesn't seem to have a github.com equivalent).

reo/scenario.py
```diff
@@ -227,7 +227,7 @@ def setup_pv(pv_dict, latitude, longitude, time_steps_per_hour):
             tmp = dict()
             tmp['federal_itc_pct'] = wind.incentives.federal.itc
             tmp['installed_cost_us_dollars_per_kw'] = wind.installed_cost_us_dollars_per_kw
-
+            print('UPDATED')
```
NLaws, line 230: looks like a debugging line that can be removed?

reo/process_results.py
TK-23: @NLaws we get capital costs from initial_capital_costs in the Financial outputs

reo/process_results.py
TK-23: @NLaws Yes, this is the LCOE for the developer. I reviewed this with Dan and he said not to use the two party factor and to use the developer discount factor which accounts for their profit

reo/process_results.py
TK-23: @NLaws good catch here! talked with Dan, he hadn't been considering tax benefits (and MACRS, federal ITC) when we spoke before about how to calculate LCOE. He's looking into it now and will have some advice on how to include these soon

reo/src/reopt_xpress_model.jl
```diff
@@ -1,9 +1,9 @@
 using JuMP
 using Xpress
 
-function reopt_model(MAXTIME)
+function reopt_model(MAXTIME, LOGFILE)
    
-    REopt = direct_model(Xpress.Optimizer(MAXTIME=-MAXTIME, OUTPUTLOG = 0))
+    REopt = direct_model(Xpress.Optimizer(MAXTIME=-MAXTIME, logfile=LOGFILE))
```
NLaws, line 6: @TK-23 We do not want the logfile enabled in any production branches (will consume too much memory).

PR #81, opened 2020-07-22
Update wind url

t-kwasnik:
## Update

Paul Edwards has stood up a new more robust wind data service - this makes use of this new url and removes legacy code


TK-23:
@sakshi-mishra can you review? thanks!

PR #88, opened 2020-07-28
Decomposition of CHP formulation

zolanaj:

This includes a decomposition of the year-long optimization model instance into month-long subproblems that can be solved in parallel

test_chp runs in 4-5 minutes on a laptop using the docker (total time) to a ~2.5% optimality gap locally on a laptop, whereas the test times out locally before getting to a 7% gap without the decomposition under the latest formulation (it was a bit faster before the hourly opex additions).  

If the procedure reaches the optimality criterion or exceeds timeout_seconds at the end of an iteration, it stops and provides the best solution found to date.  As of now, the gap is tracked but not reported.

No new inputs or imports have been made that would require a docker rebuild or migration; all the new code is in the Julia model(s) and run_jump_model.py

* **Other information**:
- The branch currently has two Julia files, which means there's a lot of redundant code.  For now, if CHP is excluded, the monolith is solved using reopt.jl.  There is code for running the monolith using reopt_decomposed.jl (where the decomposition code resides as well), and it has passed some tests (I haven't tried the full suite - yet).
- There are two bug fixes in the Julia modeling included in this PR:
79a445c334647ee24b141f44a582aa98bba1ea61 fixes an incorrectly referenced set
 acbef4253958b190704aa566be78329433ab9ce6 fixes the constraint used to model the product of a binary and continuous variable for tracking O&M costs for the CHP system as a function of size and hours run.  This was further tightened in commit 90fbbee27d96eee1a96621e6b91f7c5973d2733c and then migrated to reopt.jl in d6ddaa7255629bc217365057f8fc27818f47e2c6
- When the Xpress timeout issue is resolved, subproblem termination criteria (optimality_tolerance and timeout_seconds) should be reduced.  These are set to conservative values for now.


zolanaj:
@Bill-Becker  the idea is this: Say you have a 100kW CHP system and the max system size is 200kW.  Under the previous formulation, you could run for half of the hours and incur zero OpEx costs (per-unit-size cost is 1):
each hour running: p.OMcostPerUnitHourPerSize[t] * m[:dvSize][t] = +100
each hour not running: p.OMcostPerUnitHourPerSize[t] * m[:dvSize][t] - m[:NewMaxSize][t] * p.OMcostPerUnitHourPerSize[t] =100 - 200 = -100 

By summing for all time steps, these hours not running contribute a negative value in each hour unless the system size is the maximum allowable (since the sum is nonnegative, but not each hourly contribution).  

For what it's worth, while the OpEx constraints do add enough complexity to slow the model down, changing from the sum to the three-constraint exact linearization actually improved the performance of test_chp a little bit (at least for the decomposed problem).

Separately, I'm going to investigate failing tests today.

PR #89, opened 2020-07-29
Resilience end point

sakshi-mishra:

This PR adds new POST endpoint called resilience stats (/simjob)
The outage cost calculation does NOT happen as part of optimization process. Instead, the outage simulator module is called separately by sending  POST request to /simjob end point. 

Currently, the outage simulator is triggered automatically as soon as the optimization runs (bau and with_tech) gets done.

A new POST request needs to be sent to /simjob/ endpoint to trigger the outage simulator run (for obtaining resilience_stats data).

* **Other information**:


NLaws:
@sakshi-mishra When I test the new endpoint locally I get:
```
{
    "error_message": "Sorry, this request could not be processed. Please try again later."
}
```
I posted an optimization job, checked its status was optimal, then I POST'ed the run_uuid to `127.0.0.1:8000/v1/simjob/`. Is that the correct work flow?

I also ran your new test, which passed. However, it is printing a huge amount of data. Could you please change the test to not print anything? It also seems like the test should be in the resilience_stats/tests directory since it is testing a capability for that app: did you have a reason for placing it in reo/tests?

I wasn't able to test GET'ing the resilience stats after POST'ing my `simjob`. I noticed in the new test that it POST's to get the resilience stats. Please change the work flow to a POST/GET.

What do you think of changing `simjob` to something like `outagesimjob` ? 

We should also add some validation similar to the job endpoint for the run_uuid being a `UUID` and limiting the other parameters to valid values (such as `bau in [true, false]` or istype bool).

sakshi-mishra:
**Q1**: I posted an optimization job, checked its status was optimal, then I POST'ed the run_uuid to 127.0.0.1:8000/v1/simjob/. Is that the correct work flow?
**A1**: No. In the local development environment, the database gets cleaned up as soon as the optimization job is over. Therefore, the outage simulator is not able to fetch the scenario data attached to this UUID. The custom written test overcomes this issue by running the optimization and simulation in the same test.

**Q2** - I also ran your new test, which passed. However, it is printing a huge amount of data. Could you please change the test to not print anything? 
**A2** - I left the print statement in, my bad. Will remove it in the next commit.

**Q3**: It also seems like the test should be in the resilience_stats/tests directory since it is testing a capability for that app: did you have a reason for placing it in reo/tests?
**A3**: Since it was running optimization job first, for which reo/tests folder is responsible, I decided to keep it in reo. Happy to move it to resilienc_stats/tests.

**Q4**: I wasn't able to test GET'ing the resilience stats after POST'ing my simjob. I noticed in the new test that it POST's to get the resilience stats. Please change the work flow to a POST/GET.
**A4**: The current set up is POST and receive the stats. It returns the resilience_stats similar to how /job endpoint returns UUID. This would mean after posting the job on /simjob, UI will wait for the response. Do you see any caveat with this approach or any advantages with the POST/GET approach?

**Q5**: What do you think of changing simjob to something like outagesimjob ?
**A5**: I debated between the two and settled with simjob, happy to change it back to outagesimjob :)

**Q6**: We should also add some validation similar to the job endpoint for the run_uuid being a UUID and limiting the other parameters to valid values (such as bau in [true, false] or istype bool).
**A6**: Alright.

NLaws:
> **Q1**: I posted an optimization job, checked its status was optimal, then I POST'ed the run_uuid to 127.0.0.1:8000/v1/simjob/. Is that the correct work flow?
> **A1**: No. In the local development environment, the database gets cleaned up as soon as the optimization job is over. Therefore, the outage simulator is not able to fetch the scenario data attached to this UUID. The custom written test overcomes this issue by running the optimization and simulation in the same test.

I don't think this is true: I could not check the results status if the database were cleaned up. Have you tested the new endpoint by running local django and celery servers? That is what I was doing when I got the error message.

> **Q4**: I wasn't able to test GET'ing the resilience stats after POST'ing my simjob. I noticed in the new test that it POST's to get the resilience stats. Please change the work flow to a POST/GET.
> **A4**: The current set up is POST and receive the stats. It returns the resilience_stats similar to how /job endpoint returns UUID. This would mean after posting the job on /simjob, UI will wait for the response. Do you see any caveat with this approach or any advantages with the POST/GET approach?

We need the POST/GET approach for the new way that the UI will generate/get resilience stats. We already have the "wait" approach with the `resilience_stats` end point.

sakshi-mishra:
1. I spun up the local redis and django servers and was able to send a POST to `/outagesimjob` and receive the outage sim results back.
Here's what I did after starting the redis server, django server and the celery workers:
1. Send a POST to `/job` with the body json as in [`post_job.txt`](https://github.com/NREL/REopt_Lite_API/files/4996592/post_job.txt)
2. Got the `run_uuid` as the response:
```json
{
    "run_uuid": "6ea30f0f-3723-4fd1-8a3f-bebf8a3e4dbf"
}
```
3. Send a POST to `/outagesimjob` endpoint with the following body:
```json
{
 "run_uuid": "6ea30f0f-3723-4fd1-8a3f-bebf8a3e4dbf",
 "bau": false
}
```
4. Got the resilience results as the response as in [outagesim_response.txt](https://github.com/NREL/REopt_Lite_API/files/4996593/outagesim_response.txt)

sakshi-mishra:
@NLaws :

I have split the POST and the GET. The updated workflow is as follows:

1. User POSTs a job to `/job` & receives `run_uuid` immediately. Example response:
```json
{
    "run_uuid": "6ea30f0f-3723-4fd1-8a3f-bebf8a3e4dbf"
}
```
2. User can then POST the following data to `/outagesimjob`:

```json
{
 "run_uuid": "6ea30f0f-3723-4fd1-8a3f-bebf8a3e4dbf",
 "bau": false
}
```
It will validate the UUID, start the outage simulation task and immediately return the following response:

```json
{
    "Status": 204,
    "Success": true,
    "bau": false,
    "resource_uri": "/v1/outagesimjob/6ea30f0f-3723-4fd1-8a3f-bebf8a3e4dbf/",
    "run_uuid": "6ea30f0f-3723-4fd1-8a3f-bebf8a3e4dbf"
}
```
3. The user can then GET the resilience stats at `/job/6ea30f0f-3723-4fd1-8a3f-bebf8a3e4dbf/resilience_stats`

TK-23:
@sakshi-mishra I tried to poll the results before they were ready and got this:

{"Error": "Outage sim results not available.Please POST to /outagesimjob first to generate theresults before GET-ing the results from /resilience_stats"}


Note the URL needs  / at the end in my local deploy. Also, can you fix the typo there and make more clear what the URL and POST payload should be?

Also, when I posted {"run_uuid":"d5a3a71f-2df4-4680-9195-ff05d69b80ff"} to /outagesimjob/ I got:

{
    "error_message": "Sorry, this request could not be processed. Please try again later."
}

Here can you make more clear what is missing? Perhaps we can do default settings where bau is not set?

sakshi-mishra:
@TK-23 : Thanks for the feedback.

I have increased the verbosity of the 404 response and also have added a sample payload/body data struct to the response message.
I am committing the `/v1` from the endpoint URL as that may change.

I have set `bau` to `false` by default so the POST body in your example (without `"bau"`) should work fine now with bau set to false.

TK-23:
@sakshi-mishra thanks for adding that in! Looking really good overall. Before we deploy though, the message I got while polling for results is:

{
    "Error": "Outage sim results are not ready.Sample body data for POST-ing to /outagesimjob/: {\"run_uuid\": \"6ea30f0f-3723-4fd1-8a3f-bebf8a3e4dbf\", \"bau\": false}"
}

From looking at the code it looks like we are missing part of the message to the user. Think you need () around the multiline sting section. Also, can we make sure we have spaces after periods and that there are not escape characters in the text?

sakshi-mishra:
@TK-23 , Thanks for catching that!

I have fixed it with my last commit.

sakshi-mishra:
@NLaws  and @TK-23  - please approve and merge if all looks okay. Thank you!

NLaws:
@sakshi-mishra Thanks for addressing those issues! I tested again locally and have a couple more requested changes:

1. I like the detailed response from the new outagesimjob end point! Here is what I got:
```
{
    "Status": 204,
    "Success": true,
    "bau": false,
    "resource_uri": "/v1/outagesimjob/9f3192d5-fd3e-4586-9963-0c8180a0f40c/",
    "run_uuid": "9f3192d5-fd3e-4586-9963-0c8180a0f40c"
}
```
    Can you please make the Status 201 when the POST is successful ("created"). 204 is meant for PUT's. 
2. Since the response from the new outagsimjob POST endpoint includes the resource URI I clicked on it expecting to get my outage simulator results - but I get a 405 (method not allowed) response. I think that we should make the outagesimjob behavior match that of the job endpoint, where one can GET the results of the POST'ed job at outagesimjob/UUID/results. What do you think? It looks the set up you've made expects the user to GET the outagesimjob results at the existing job/UUID/resilience_stats endpoint right? I think that we should keep that endpoint for backwards compatibility as well as providing the option to GET the outage sim results as a blocking process for those that want it.

TK-23:
@sakshi-mishra just approved in advance of you addressing Nick's comments

sakshi-mishra:
Hi @NLaws - thanks for the feedback.

item 1 - waiting for Andrew J. to confirm (either 201 or 204)
item 2 - is this a must-have-today before merging request or can we merge without this functionality and do another PR in a day or two to add this nice-to-have behavior?

Trying to put this PR (with required functionality in place for UI to work with it) on fast track for the `chpbeta` merge...

sakshi-mishra:
@NLaws  - fixed item #1 (my VM time wasn't correct so recent commits are showing up 14 hours ago! ). Please let me know if item #2 above is a blocker for the merge. Thank you!

sakshi-mishra:
@NLaws  - issue is created and the base branch is changed to chpbeta. 

I will go ahead and merge it. Thanks.

PR #94, opened 2020-08-05
Increase Julia threads to 2

NLaws:
@Bill-Becker @t-kwasnik I'm testing this out on c110p now using a CHP scenario on chpbeta branch. Note that this in effect determines how many cores each celery worker can use because the celery workers' most computationally intensive task is solving the JuMP model.  We should try it out with more threads once the `parallelsubproblems` branch is ready for deployment. Keep in mind each server has 16 cores and we can't control how many jobs are POST'ed at a time - so we probably should be conservative with the settings.  @zolanaj as you are testing locally you could increase your JULIA_NUM_THREADS too (maybe you already have). @t-kwasnik this could be something to evaluate in your benchmarking as well.

Note that this change will set the value for any server that gets this change deployed to it. If we need to we could set server specific JULIA_NUM_THREADS as well.


NLaws:
This leads to the celery worker memory growth issue on c110p, making a CHP problem run slower and slower over time until the OOM killer takes over.

PR #103, opened 2020-08-20
added curtailed wind output

t-kwasnik:
## New

Track exported wind electricity during a grid outage as curtailed


t-kwasnik:
@sakshi-mishra can you take a look at this after the test suite completes?

PR #109, opened 2020-09-01
Improve Lookback peak demand model

NLaws:
With URDB's upgrade to v8 lookback demand ratchets have been improved. Now the response keys allow accounting for seasonal lookbacks and rolling lookback ranges:
```
URDB lookback fields:
            lookbackMonths	
            Type: array
            Array of 12 booleans, true or false, indicating months in which lookbackPercent applies. 
                If any of these is true, lookbackRange should be zero.
            
            lookbackPercent	
            Type: decimal
            Lookback percentage. Applies to either lookbackMonths with value=1, or a lookbackRange.
            
            lookbackRange	
            Type: integer
            Number of months for which lookbackPercent applies. If not 0, lookbackMonths values should all be 0.
```

Prior to this change REopt Lite and URDB only modeled the seasonal lookback. Now we model a rolling lookback if `lookbackRange` is non-zero.


reo/src/reopt.jl
```diff
@@ -644,16 +644,48 @@ function add_tou_demand_charge_constraints(m, p)
...
-	)
+	if p.DemandLookbackRange != 0  # then the dvPeakDemandELookback varies by month
+		
+		##Constraint (12e): dvPeakDemandELookback is the highest peak demand in DemandLookbackMonths
+		for mth in p.Month
+			if p.Month > p.DemandLookbackRange
+				@constraint(m, [lm in 1:p.DemandLookbackRange],
+					m[:dvPeakDemandELookback][mth] >= 
+					sum(m[:dvPeakDemandEMonth][mth - lm, n] for n in p.DemandMonthsBin)
+				)
```
kkrah, line 655: Instead of setting the dvPeakDemandELookback for a given month based on the relevant lookback months' dvPeakDemandEMonth, might it make sense to set it based on the lookback months' peak dvGridPurchase (might require a new variable, e.g. dvPeakDemandEMonthActual)? My concern is that using dvPeakDemandEMonth, peaks will be propogated through the whole year.

reo/src/reopt.jl
```diff
@@ -644,16 +644,48 @@ function add_tou_demand_charge_constraints(m, p)
...
+				for lm in 1:p.DemandLookbackRange
+					lkbkmonth = mth - lm
+					if lkbkmonth ≤ 0
+						lkbkmonth += 12
+					end
+					@constraint(m,
+						m[:dvPeakDemandELookback][mth] >= 
+						sum(m[:dvPeakDemandEMonth][lkbkmonth, n] for n in p.DemandMonthsBin)
+					)
+				end
```
kkrah, line 666: Would the logic under "else" work for all months, to avoid the if-else? (If breaking this into if and else speeds up code, please ignore this comment!)

reo/src/reopt.jl
kkrah: For example: Let's say we have a 3-month rolling lookback of 100%. In August, dvPeakDemandELookback (August) and therefore dvPeakDemandEMonth(August) would be set based on May-July and August. So if the June peak is higher than the August peak, dvPeakDemandEMonth(August) would be set based on the June peak (which is what we want). However, then let's look at October.  dvPeakDemandELookback (October) and therefore dvPeakDemandEMonth(October) should be set based on July-September and October. However, since dvPeakDemandEMonth(August) is being set by the June peak, if dvPeakDemandELookback (October) is set based on dvPeakDemandEMonth then June's peak would be propagated through to October, even though June is more than 3 months before October.

reo/src/reopt.jl
```diff
@@ -644,16 +644,48 @@ function add_tou_demand_charge_constraints(m, p)
...
-	##Constraint (12f): Ratchet peak demand charge is bounded below by lookback
-	@constraint(m, [r in p.DemandLookbackMonths],
-		sum( m[:dvPeakDemandEMonth][r,e] for e in p.DemandBin ) >= 
-		p.DemandLookbackPercent * m[:dvPeakDemandELookback] 
-	)
+	if p.DemandLookbackRange != 0  # then the dvPeakDemandELookback varies by month
+		
+		##Constraint (12e): dvPeakDemandELookback is the highest peak demand in DemandLookbackMonths
+		for mth in p.Month
+			if p.Month > p.DemandLookbackRange
```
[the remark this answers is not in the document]
NLaws: Yes

reo/src/reopt.jl
NLaws: Yes that makes sense. I just committed a change using `dvGridPurchase`, thank you!

reo/src/reopt.jl
NLaws: the if-else is to handle lookback months into the "previous" year (eg. month = 1, lookback = 3), I couldn't think of way to handle it in one block

NLaws: @kkrah and it might speed things up having the vectorized constraints in the `if` block.

PR #110, opened 2020-09-02
Fix export during outages

NLaws:
# Bug fix
- DER export to grid (in NEM and wholesale `SalesTiers`) was not set to zero during `TimeStepsWithoutGrid`.

# Minor updates
- removed obsolete `tech_to_tech_class` from data_manager

### DIFF
--- a/reo/src/reopt.jl
+++ b/reo/src/reopt.jl
@@ -573,11 +573,26 @@ end
 
 
 function add_nem_constraint(m, p)
+	# NEM is SalesTier 1
+	# dvStorageToGrid is always fixed at 0.0, remove it?
 	@constraint(m, GridSalesLimit, 
 		p.TimeStepScaling * sum(m[:dvProductionToGrid][t,1,ts] for t in p.TechsBySalesTier[1], ts in p.TimeStep)  + 
-		sum(m[:dvStorageToGrid][u,ts] for u in p.StorageSalesTiers, ts in p.TimeStep) <= p.TimeStepScaling * 
-		sum(m[:dvGridPurchase][u,ts] for u in p.PricingTier, ts in p.TimeStep)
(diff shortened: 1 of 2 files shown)


reo/src/reopt.jl
```diff
@@ -573,14 +573,29 @@ end
 
 
 function add_nem_constraint(m, p)
+	# NEM is SalesTier 1
+	# dvStorageToGrid is always fixed at 0.0, remove it?
```
zolanaj, line 577: Given the previous conversation in which storage is only intended to meet site load, it can probably be removed here and elsewhere, e.g., the site load constraint (8a).  It might make sense to do that in a separate PR though (I can do it if you prefer).

reo/src/reopt.jl
```diff
@@ -757,6 +772,7 @@ function reopt_run(m, p::Parameter)
 	## Temporary workaround for outages TimeStepsWithoutGrid
 	if !isempty(p.TimeStepsWithoutGrid)
 		add_no_grid_constraints(m, p)
+		add_no_grid_export_constraint(m, p)
```
zolanaj, line 775: would it make sense to combine add_no_grid_constraints and add_no_grid_export_constraints into one function if they're only ever together?

zolanaj (approved):
Looks good and can be merged as is, with minor comments attached (one of which might warrant further discussion)

reo/src/reopt.jl
NLaws: I put them in separate functions because one of the speed tips from Oscar Dowson was to add model constraints in individual functions.

reo/src/reopt.jl
NLaws: Yes sounds good, but given the CHP merge in the pipeline I suggest we wait to reduce the number of conflicts when we merge CHP.

NLaws: Or you could remove them from both develop and chpbeta (and parallelsubproblems)

PR #112, opened 2020-09-09
CHP cost curve

Bill-Becker:
- [] Docs have been added / updated (for bug fixes / features)
a. Changes just apply to CHP for now, and I will update the documentation from the AMO project for this

Feature, with cost curve framework for other technologies but only implemented currently for CHP.

For each tech, `installed_cost_us_dollars_per_kw` is a scalar value and the model therefore cannot account for economy of scale for technology capital costs

1. Add framework for inputting a cost curve, where `installed_cost_us_dollars_per_kw` can be input as a list if there is a corresponding `tech_size_for_cost_curve` input of a list with the same length as `installed_cost_us_dollars_per_kw`.
a. This new `tech_size_for_cost_curve` input was only added for CHP, so this input would need to be added for other technologies to have the same functionality
2. Add `tech_size_for_cost_curve` input for CHP tech cost curve
3. Update default values for CHP size classes which now uses the cost curve input functionality. 

No, backwards compatibility of inputting only a single scalar value for `installed_cost_us_dollars_per_kw` has been maintained

* **Other information**:
1. New test is in test_chp.py called test_cost_curve
2. Added inputs validation to make sure the length of `installed_cost_us_dollars_per_kw` and `tech_size_for_cost_curve` are the same, and that the `tech_size_for_cost_curve` is in ascending order.
3. `installed_cost_us_dollars_per_kw` for CHP is now converted to a list even if it's a scalar user input (same as other inputs with type ["float", "list_of_float"])
3. This new feature and 3-segment cost curve default for CHP was benchmark tested using the Monolith JuMP model, and there was no noticeable impact on solve times


Bill-Becker:
@zolanaj - I'm adding you as a reviewer in case this has any consequences in the decomposition model, but I don't think it does. You don't have to thoroughly review the pre-processing side of this cost curve as an input.

@NLaws Please review this cost curve input functionality for CHP. I also tried to make it so that it will be easy to add this functionality for other techs like PV and Wind in the future. I believe it is working in conjunction with the capital cost incentives processing, and I tested that for a few different combinations of incentives.

zolanaj (approved):
Since there are no changes to reopt.jl in this pull request, there should be nothing to edit in the decomposition (via reopt_decomposed.jl)

Bill-Becker:
@NLaws Do you have a guestimate as to when you could review this? If you're overloaded and think someone else is suitable to review, please advise. I'd like to merge this soon for the AMO project deliverable.

reo/validators.py
```diff
@@ -1025,6 +1039,22 @@ def check_special_cases(self, object_name_path, template_values=None, real_value
                         if filtered_values['min_allowable_kw'] or 0 > filtered_values['max_kw'] or 0:
                             self.input_data_errors.append('The CHP min_allowable_kw cannot be greater than its max_kw')
 
+                        # Cost curve
+                        if len(filtered_values['installed_cost_us_dollars_per_kw']) > 1:
```
NLaws, line 1043: @Bill-Becker this new block of code looks exactly like the new block on lines 986-999. Could it be a sub-function that is called twice?

reo/tests/test_chp.py
```diff
@@ -61,3 +61,63 @@ def test_chp_sizing(self):
             print("Error message: {}".format(d['messages'].get('error')))
             raise
 
+    def test_cost_curve(self):
```
NLaws, line 64: @Bill-Becker As you mentioned in the commit summary this new CHP cost curve capability did not make any existing tests slower. That is to be expected since no existing tests use the new capability. It would be interesting to know if you used the cost curve in `test_chp_sizing` what impact it has on the solve time.  If it doesn't significantly impact that test's time then we could integrate your new test with the existing one so that there is little net impact on the time it takes for the total test suite to run. Just a suggestion and not necessary.

reo/tests/test_chp.py
```diff
@@ -61,3 +61,63 @@ def test_chp_sizing(self):
...
+        nested_data["Scenario"]["Site"]["CHP"]["tech_size_for_cost_curve"] = [100, 600, 1140]
+        #nested_data["Scenario"]["Site"]["CHP"]["utility_rebate_us_dollars_per_kw"] = 50
+        #nested_data["Scenario"]["Site"]["CHP"]["utility_rebate_max_us_dollars"] = 1.0E10 #50 * 200
+        #nested_data["Scenario"]["Site"]["CHP"]["federal_rebate_us_dollars_per_kw"] = 50
+        nested_data["Scenario"]["Site"]["CHP"]["federal_itc_pct"] = 0.1
+        nested_data["Scenario"]["Site"]["CHP"]["macrs_option_years"] = 0
+        nested_data["Scenario"]["Site"]["CHP"]["macrs_bonus_pct"] = 0
+        nested_data["Scenario"]["Site"]["CHP"]["macrs_itc_reduction"] = 0.0
+
+        init_capex_chp_expected = 2020666.67
```
NLaws, line 91: @Bill-Becker could you please add a comment for how you arrived at this expected value? (I tried the slope * size but there must be a y-intercept?)

NLaws (approved):
See inline comments for some suggested changes.

reo/tests/test_chp.py
Bill-Becker: @NLaws The testing I did was the ~120x scenario benchmark testing using the monolith (which we're going with for the baseline model until we improve decomposition), and there was an overall positive benefit to time/non-timeout/non-negative-NPV. I documented this in the benchmark testing doc on Box. Good idea to combine CHP tests as this will be the new default cost input for CHP.

reo/validators.py
Bill-Becker: Addressed in this [commit].(https://github.com/NREL/REopt_Lite_API/pull/112/commits/43495dadee6dfdb37a4e664f588da540af7b6a04)

reo/tests/test_chp.py
Bill-Becker: Addressed in this [commit](https://github.com/NREL/REopt_Lite_API/pull/112/commits/466a2f9d85a91e0ec39b58707fae4cf6459bb5d5)

PR #115, opened 2020-09-16
Urdb parse

t-kwasnik:
## Bug Fix

adding line to validators so that the URDB_RateValidator sees rates returned from the openei API

## New
require energyratestructure, energyweekendschedule and energyweekdayschedule in a URDB rate
added new checking of URDB float fields

### DIFF
--- a/reo/validators.py
+++ b/reo/validators.py
@@ -50,7 +50,7 @@ class URDB_RateValidator:
 
     error_folder = 'urdb_rate_errors'
 
-    def __init__(self,_log_errors=True, **kwargs):
+    def __init__(self, _log_errors=True, **kwargs):
         """
         Takes a dictionary parsed from a URDB Rate Json response
         - See http://en.openei.org/services/doc/rest/util_rates/?version=3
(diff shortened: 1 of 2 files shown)


reo/validators.py
```diff
@@ -246,6 +284,26 @@ def recursive_search(item,level=0, entry=0):
...
+    def isNotNone(self, name):
+        if getattr(self,name, None) is None:
+            self.errors.append('Missing valid entry for {}.'.format(name))
+            return False
+        return True
+    
+    def isNotEmptyList(self, name):
+        if type(getattr(self,name)) != list:
+            self.errors.append('Expecting a list for {}.'.format(name))
+            if len(getattr(self,name)) == 0:
```
NLaws, line 304: should the `if` for the length check be de-dented? (Unless it self.name is a list it probably doesn't have a length).

reo/validators.py
```diff
@@ -246,6 +284,26 @@ def recursive_search(item,level=0, entry=0):
...
+                self.errors.append('List is empty for {}.'.format(name))
+
```
NLaws, line 306: Do you want `isNotEmptyList` to return `None` or should it return `True/False` like `isNotNone`?

NLaws (approved):
approved pending CI

PR #116, opened 2020-09-16
200916 master<-dev improved lookback peak demand, URDB v8

NLaws:
# Improved Lookback Peak Demand model
With URDB's upgrade to v8 lookback demand ratchets have been improved. Now the response keys allow accounting for seasonal lookbacks and rolling lookback ranges:
```
URDB lookback fields:
            lookbackMonths	
            Type: array
            Array of 12 booleans, true or false, indicating months in which lookbackPercent applies. 
                If any of these is true, lookbackRange should be zero.
            
            lookbackPercent	
            Type: decimal
            Lookback percentage. Applies to either lookbackMonths with value=1, or a lookbackRange.
            
            lookbackRange	
            Type: integer
            Number of months for which lookbackPercent applies. If not 0, lookbackMonths values should all be 0.
```

Prior to this change REopt Lite and URDB only modeled the seasonal lookback. Now we model a rolling lookback if `lookbackRange` is non-zero.

# Bug Fixes
Change NaN to null for irr_pct


NLaws:
@t-kwasnik Kathleen already reviewed the lookback peak demand model in the julia code. I would like your check on the URDB changes and also for your awareness (this PR upgrades the URDB to v8).

t-kwasnik (approved):

PR #117, opened 2020-09-17
fix pv input to outagesim

t-kwasnik:
## bug fix

addresses multiple pvs and a division by 0 case in outage simulator inputs


t-kwasnik:
@sakshi-mishra can you take a look at this fix to  the outage simulator inputs? I was getting a division by zero error in the BAU case when the pv size was zero, and also noticed it wasn't set up to account for all pvs

sakshi-mishra (approved):
Thanks for catching and fixing this issue, @t-kwasnik ! Changes look good. Thanks!

PR #118, opened 2020-09-17
Add Optimization Parameter Inputs; Default to Monolith for CHP

Bill-Becker:

Allows additional customization of optimization model parameters. 

1. Can only specify one optimality tolerance that gets applied to both bau and non-bau scenarios
2. Can only use decomposition for cases considering CHP
3. Cannot specify decomposition subproblem optimality tolerance (hardcoded in reopt_decomposed.jl)

1. Can specify optimality tolerance for:
a. bau
b. non-bau (with techs considered)
c. decomposition subproblems
2. Can choose whether or not to use the decomposition for the non-bau (with techs) scenario (bau scenario always uses monolith)
3. By default, the monolith will be used for all scenarios (even CHP!)
4. Changed the minimum values for all optimality_tolerance_ inputs to be 1e-5
5. Increased the maximum value for `timeout_seconds` to 5000 (kept default at 420).

Kinda/maybe; if users are currently specifying `optimality_tolerance` (webtool does this), then that needs to be broken up into `optimality_tolerance_bau` and `optimality_tolerance_techs`

* **Other information**:


reo/src/reopt_decomposed.jl
```diff
@@ -883,7 +883,7 @@ end
 
 function add_decomp_model(m, p::Parameter, model_type::String, mth::Int64)
 	if m[:solver_name] == "Xpress"
-		sub_model = direct_model(Xpress.Optimizer(MAXTIME=-120, MIPRELSTOP=0.03, OUTPUTLOG = 0))
+		sub_model = direct_model(Xpress.Optimizer(MAXTIME=-120, MIPRELSTOP=p.DecompOptTol, OUTPUTLOG = 0))
```
zolanaj, line 886: For completeness, we should update ratioGap for Cbc and limits_gap for SCIP too.  (Likewise for MAXTIME, seconds, and limits_time if we include time limits as a user-defined option as suggested above.

reo/nested_inputs.py
```diff
@@ -109,13 +109,32 @@ def list_of_str(input):
...
       "description": "The threshold for the difference between the solution's objective value and the best possible value at which the solver terminates"
     },
+	"optimality_tolerance_techs": {
+      "type": "float",
+      "min": 1.0e-5,
+      "max": 10.0,
+      "default": 0.001,
+      "description": "The threshold for the difference between the solution's objective value and the best possible value at which the solver terminates"
+    },
+	"use_decomposition_model": {
```
zolanaj, line 126: If we're going to allow the user to toggle the decomposition model on/off and specify the subproblem gap, I think we should also allow the subproblem time limit to be customized too.  Would you agree?

zolanaj (approved):
Thanks for doing this, and I agree in general with the approach.  One minor comment and one bigger comment, which I'd like your feedback on but can address in a separate PR if preferred.

reo/src/reopt_decomposed.jl
Bill-Becker: @zolanaj Can you explain if `ratioGap` and `limits_gap` are identical/equivalent to `MIPRELSTOP`, and if so we can just assign those with `p.DecompOptTol`? Relatedly with `MAXTIME` (which I agree we should have that as an input too for the subproblems), is `seconds` and `limits_time` equivalent?

zolanaj: Yes,  `ratioGap` and `limits_gap` are the Cbc and Scip analogs to MIPRELSTOP and can be set using p.DecompOptTol.  Likewise for `seconds` and `limits_time` being analogous to MAXTIME for the other solvers; a single parameter should be able to populate these with the same input (e.g., `p.DecompSubTimeLimit`, or a different name if you prefer).

Issue #77, opened 2020-07-13
Julia throwing exception when optimization times out in Docker


zolanaj:
**Note: for support questions and enhancement suggestions please use the REopt Forum (under development)**. This repository's issues are reserved for feature requests and bug reports.

* **I'm submitting a ...**
  - [x] bug report

* **What is the current behavior?**
When Xpress times out (i.e., fails to acheive the desired optimality gap in the timeframe provided as input (timeout_seconds), Julia raises  DimensionMismatch error within the optimize!(m) call.  This occurs even when there are feasible solutions found (per the Julia logfiles).  The error message is as follows: 
exc_type: <class 'RuntimeError'>; exc_value: Julia exception: DimensionMismatch("array could not be broadcast to match destination")

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via
https://plnkr.co or similar (you can use this template as a starting point: http://plnkr.co/edit/tpl:AvJOMERrnz94ekVua0u5).
in the current version of 'chp-reformulation': run test_heating_and_cooling_load_profiles out of the docker.
in the current version of 'develop':
1) in reo\tests\posts\critical_load_bau_can_sustain_outage.json, change the value of "timeout_seconds" to 1
2) out of the docker container, run "docker exec -it celery python manage.py test reo.tests.test_critical_load_bau"

* **What is the expected behavior?**
After a timeout, either a feasible solution should be returned if one is found, or outputs should be generated without a feasible solution populated.  

* **What is the motivation / use case for changing the behavior?**
For the decomposition solution method, it is likely that some subproblems will not solve to the desired optimality gap in the allotted timeframe, and to ensure high performance of the algorithm, it is preferable to have this in place.  More complex instances (e.g., those including CHP, storage, and multi-tiered TOU charges) are likely to reach a timeout when solved directly as well; in thee cases, it is best to provide the best solution found within the timeframe.

* **Please tell us about your environment:**
  
  - Current build of the 'develop' branch, out of the docker container

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)
Full stacktrace for the test_critical_load_bau instance: 
Stacktrace:
 [1] check_broadcast_shape at ./broadcast.jl:508 [inlined]
 [2] check_broadcast_axes at ./broadcast.jl:511 [inlined]
[...]

PR #119, opened 2020-09-18
Fix timeout bug - close Issue #77

zolanaj:

Fixes Issue #77: there were two bugs in Xpress.jl which (i) outputted the objective value when the best bound was queried, and (ii) returned a dimensionmismatch error when the optimization model timed out (but a feasible solution was found).  These issues have been fixed, and this PR updates the reference to the Xpress library and makes a small update to the exception handling so that timeouts return a solution instead of raising an exception.

https://github.com/NREL/REopt_Lite_API/issues/77

Lower bounds are reported properly, and the best feasible solution is returned if the model times out.

No

* **Other information**:


NLaws:
@zolanaj what happens with a timeout and when no solution has been found? Also, I'm not sure that we want to report sub-optimal solutions, just that the problem timed out before an optimal solution has been found. Users are expecting only optimal solutions from the API/Webtool - and not feasible solutions. Will these changes report null values for the outputs and a message that the problem timed-out? (That is the behavior that users are expecting).

Bill-Becker:
@NLaws @zolanaj Good questions, and I agree we don't want to report suboptimal (higher than specified optimality tolerance) solutions especially in light of the negative NPV solutions we've been seeing. However, I think it would be useful to **allow** an API user to choose whether they'd like to see the best feasible solution to optimization found within the specified timeout limit which would be useful if the actual gap was reported along with it. Unfortunately I think Xpress.jl is still not reporting the gap, so maybe we punt on that until that issue is resolved. For the web tool, we'll need to just make sure we're delivering "optimal" solutions (within the web tool's specified optimality tolerance which is TBD for CHP).

zolanaj:
@NLaws @Bill-Becker  Thanks for the feedback.

I will revert the change that outputs a suboptimal solution in the event of a timeout when the monolith is solved.  That way the tool will continue to work as expected, and we can still use this feature when subproblems time out in the decomposition (which was hindered by the bug in the first place).  If we think collectively that suboptimal solutions may be worth presenting (e.g., if they have a positive NPV), then I can add a user-defined option for that, but I'd prefer to do that in a separate PR.

When I test this using test_chp, solving directly (i.e., running reopt.jl) and setting the time limit to 60 seconds and optmality tolerance to 0.01, I get the following: 
results["lcc"]:  1.3476072e7
results["lower_bound"]:  1.3150909e7
results["optimality_gap"]:  0.024128884202038577

I did have to rebuild to get this to work, that might be the difference.

Setting the time limit to one second yields no solution and now raises the same Optimization timeout we had before (and in the example above with 60 seconds, the same exception is raised).  So, the experience when solving the model directly is the same, and we now have (1) subproblem solutions returned after timeouts, and (2) more accurate bounds tracking out of JuMP when we run the decomposition.

reo/src/run_jump_model.py
```diff
@@ -193,10 +193,6 @@ def run_jump_model(self, dfm, data, run_uuid, bau=False):
     except Exception as e:
         if isinstance(e, REoptFailedToStartError):
             raise e
-        elif "DimensionMismatch" in e.args[0]:  # bug in Xpress.jl and/or JuMP that mishandles timeouts
```
NLaws, line 196: @zolanaj it seems like we still need this block? I just ran a scenario on this branch with a 1 second timeout and the celery logs still show the `DimensionMismatch` error and the API status for results is `"status": "An error occurred. See messages for more."`

zolanaj:
@nlaws I'm addressing this via commit 705f61e1b3d1e2a0a3be3720926d1918273d5e2b as I believe this may still happen if no feasible solution has been found (someone on the JuMP team is looking into this).  However when I ran a one-second model on the docker locally, an OptimizationTimeout error is returned as expected with the message  " Optimization exceeded timeout: 1 seconds. Please email reopt@nrel.gov with your run_uuid (563ee01f-1c0e-471a-bab7-e7ff3888ec6b) for support."  Is it possible that a rebuild of the image is required since it's an external library version change?

NLaws:
@zolanaj before https://github.com/NREL/REopt_Lite_API/commit/705f61e1b3d1e2a0a3be3720926d1918273d5e2b the `elif "DimensionMismatch" in e.args[0]` had been removed (in  https://github.com/NREL/REopt_Lite_API/pull/119/commits/0762b65bd7d3b30a74c017da27157b4b15237ea3), so with a time out one would get an `UnexpectedError`

zolanaj:
@NLaws I thought the Xpress library update would have removed that, and it would then be caught by the "timed-out" status returned by the results dictionary in reopt.jl instead, and the check for status == "timed-out" in the new code in 0762b65bd7d3b30a74c017da27157b4b15237ea3 would then lead to an OptimizationTimeout error (whereas formerly, it would never get to that point, because the DimensionMismatch error would prevent anything of use from being reported in the results dictionary).  With 705f61e1b3d1e2a0a3be3720926d1918273d5e2b restoring the original snippet you referenced, we're also covered if the DimensionMismatch error comes back, but this update also helps us analyze solutions if he model times out (helpful for the decomposition, but also potentially helpful if the optimality gap isn't achieved but a positive-NPV solution is found).

Bill-Becker:
@NLaws @zolanaj I was trying to re-read the comments and see if there were any open items before merging this into chpbeta. If I understand correctly, the current behavior for the monolith is the same as before where the user gets a timeout error if either the model finds a feasible non-optimal solution or no solution. Is that right? If so, @NLaws do you approve this?

@zolanaj Is the optimality gap now reported (correctly) for the monolith model? (it was an output field before, but Xpress.jl wasn't correctly assigning the LB).

I like the idea of adding an option to pass a non-optimal solution for benchmarking and comparison of monolith to decomposition. We can make that a next step beyond this PR.

zolanaj:
@Bill-Becker I can confirm that the lower bound is accurate when using Xpress on this branch.  It matches the output reported from the log file (an earlier comment show an example from test_chp with a 60-second limit, run locally from the docker).

NLaws:
@Bill-Becker yes

NLaws (approved):

Bill-Becker (approved):

PR #120, opened 2020-09-21
Fix fuelcosts

NLaws:
    - existing tests suffice
     - this PR is a bug fix, code clarification, and removal of obsolete code

## Changes
- Bug fix: Diesel fuel costs were indexed on electric tariff tiers, which was necessary before the reformulation, but now leads to an index error in the JuMP model.
- Many obsolete values are removed from `UrdbParse`
- For clarity: Moved creation of Generator specific (fuel burning tech) JuMP model arrays from `UrdbParse` to `DataManager` (because they no longer need to be created in `UrdbParse` since reformulation removed the unnecessary indices).

## Other information
When merging `chpbeta` into `develop` or `master` we will need to move [this code](https://github.com/NREL/REopt_Lite_API/blob/afe0a200f67d930ee0a972ffd212659664455a61/reo/src/urdb_parse.py#L533) as well as the [fuel burn rate/intercept code](https://github.com/NREL/REopt_Lite_API/blob/afe0a200f67d930ee0a972ffd212659664455a61/reo/src/urdb_parse.py#L578) to the new method created in this PR in the DataManager [here](https://github.com/NREL/REopt_Lite_API/blob/43307a7a93f063f51eeff4afdcdbb8d4f93079f8/reo/src/data_manager.py#L749)


Bill-Becker (approved):
@NLaws This looks good and is the right way to do it, I'll just have to convert CHP+ to this.

PR #124, opened 2020-09-23
Chpbeta move fuel params

Bill-Becker:

Moved all fuel burn parameters out of urdb_parse and into a separate file/class to be called by data_manager.

Nick moved diesel fuel burn in develop, and I made a further change to consolidate all fuel burn parameters, including CHP and boiler, into a separate file fuel_params.py which has a FuelParams class.

No, but there are 2 tests that I know of that are not passing which I believe is just due to the recent merging of develop into chpbeta:
1. test_tiered_energy_rate_and_profiler
2. test_curtailment

* **Other information**:


Bill-Becker:
@zolanaj @NLaws Note the two tests that I mention in the PR description are not passing, but I believe that is due to the recent merge of develop into chpbeta. This branch was branched off of chpbeta after I merged develop but had not fully resolved all tests that may not be related to the main change of this PR. Please let me know if you are able to help debug these two tests which are related to tiered rates and curtailment.

reo/src/fuel_params.py
```diff
@@ -0,0 +1,131 @@
...
+        self.boiler = boiler
+
+        # Assign monthly fuel rates for boiler and chp and then convert to timestep intervals
+        self.boiler_fuel_blended_monthly_rates_us_dollars_per_mmbtu = fuel_tariff.monthly_rates('boiler')
+        self.chp_fuel_blended_monthly_rates_us_dollars_per_mmbtu = fuel_tariff.monthly_rates('chp')
+        self.chp_fuel_rate_array = []
+        self.boiler_fuel_rate_array = []
+        for month in range(0, 12):
+            # Create full length (timestep) array of NG cost in $/MMBtu
+            self.boiler_fuel_rate_array.extend([self.boiler_fuel_blended_monthly_rates_us_dollars_per_mmbtu[month]] *
```
NLaws, line 62: When I moved the Generator parameters into the DataManager from UrdbParse I noticed that there was no need for creating these arrays of fuel costs (just repeating the same value and no time index in the JuMP model). Is that also true for CHP fuel costs? If so we don't need these arrays.

Bill-Becker: @NLaws for CHP and boiler, the fuel price can vary by month. So we could just feed in an array of 12 to JuMP if we have a [TimeStepsByMonth] parameter to align the 12 months to each timestep. For Diesel, I agree it only needs a single value. Just confirming, the suggestion is that we downsample the index for FuelCost in the JuMP model, right?

reo/src/data_manager.py
```diff
@@ -1202,13 +1175,13 @@ def finalize(self):
...
 
-        #TODO: handle chp parameter population directly in urdb_parse, or move all to data_manager
-        chp_thermal_prod_slope = tariff_args.chp_thermal_prod_slope if len(chp_techs) > 0 else []
-        chp_thermal_prod_slope_bau = tariff_args.chp_thermal_prod_slope_bau if len(chp_techs_bau) > 0 else []
-        chp_thermal_prod_intercept = tariff_args.chp_thermal_prod_intercept if len(chp_techs) > 0 else []
-        chp_thermal_prod_intercept_bau = tariff_args.chp_thermal_prod_intercept_bau if len(chp_techs_bau) > 0 else []
-        chp_fuel_burn_intercept = tariff_args.chp_fuel_burn_intercept if len(chp_techs) > 0 else []
-        chp_fuel_burn_intercept_bau = tariff_args.chp_fuel_burn_intercept_bau if len(chp_techs_bau) > 0 else []
+        # Specific attributes for CHP tech
+        chp_thermal_prod_slope = fuel_params.chp_thermal_prod_slope if len(chp_techs) > 0 else []
```
NLaws, line 1179: could this logic be handled in FuelParams? We could also pass fuel burning techs into FuelParams, `fb_techs` on the develop branch, instead of all techs

reo/src/fuel_params.py
NLaws: @Bill-Becker I see. On develop `d["FuelCost"] = AxisArray(d["FuelCost"], d["FuelType"])` but on chpbeta `d["FuelCost"] = vector_to_axisarray(d["FuelCost"], d["FuelType"], d[:TimeStep])`. But we only need a month index, then set up the constraints to use `TimeStepRatchetsMonth` (which I think that we can rename to `TimeStepsMonthly`?).

NLaws: (or `TimeStepsByMonth` as you suggested)

NLaws:
Regarding `test_tiered_energy_rate_and_profiler`, I ran it locally and got an error in the DataManager for too many values to unpack from `self._get_REopt_production_incentives`, which I fixed and pushed up. Now the problem is:

`AssertionError: 30576.41 != 2342.88 : Year one energy bill (30576.41) does not equal expected cost (2342.88).`

I don't know why the year one bill got so much greater. I did notice that the inputs use a URDB rate name (and utility) rather than a urdb_label. All tests that depend on tariffs should use the urdb_label to tie the test to a certain version of the tariff (otherwise the tariff could be updated in the URDB  after the test is defined - and using the rate+utility name will return the latest version of the tariff).

zolanaj:
For what it's worth, test_curtailment is also dealing with a very different objective value, but I think it may be tied to defaults since the tariff uses a urdb label.  I'm investigating the solution differences between the branches now - will follow up with an updated test.

Bill-Becker:
> Regarding `test_tiered_energy_rate_and_profiler`, I ran it locally and got an error in the DataManager for too many values to unpack from `self._get_REopt_production_incentives`, which I fixed and pushed up (Note that this problem must have been causing most reo tests to fail).
> 
> Now the problem is:
> 
> `AssertionError: 30576.41 != 2342.88 : Year one energy bill (30576.41) does not equal expected cost (2342.88).`
> 
> I don't know why the year one bill got so much greater. I did notice that the inputs use a URDB rate name (and utility) rather than a urdb_label. All tests that depend on tariffs should use the urdb_label to tie the test to a certain version of the tariff (otherwise the tariff could be updated in the URDB after the test is defined - and using the rate+utility name will return the latest version of the tariff).

@NLaws yeah I had pushed a fix for the `self._get_REopt_production_incentives` unpacking issue to chpbeta just before branching off to this one, so I was getting the same error with the energy bill as your mentioning here.

reo/src/data_manager.py
Bill-Becker: Yes, I made this update in [this commit](https://github.com/NREL/REopt_Lite_API/pull/124/commits/442d831a012daebf8280d1e47df4fd7479544cf3)

zolanaj:
update: I updated test_curtailment to include a "hard-coded" load profile that I heisted from the develop branch test, rather than taking a default profile (that turned out to be different between branches).  I also added more inputs to the test post (dsabling generator and CHP, added battery parameters).  Values are slightly different but checking the power flows before and durign the outage showed that they are appropriate; my guess is the difference is either in the load profile translation or the battery deafult values. test_curtailment is passing now.

zolanaj (approved):

Bill-Becker:
@NLaws how would you like to move forward to fix the issue with test_tiered...?

PR #132, opened 2020-10-28
201028 master<-develop

NLaws:
## Bug Fixes
- remove third-party factor from `initial_capital_costs_after_incentives` output
- convert erroneous wind radius to wind diameters
- handle MACRS when analysis years is less than 6 in results processing and proforma
- parse URDB rates to make sure at least an energy rate is supplied
- fix OutOfMemory error in docker containers when building constraints in models that have more than one time step per hour

## Improvements
- create Dockerfile.dev for Kubernetes deploys


danolis (commented):
Looks good.

PR #147, opened 2020-11-12
update validation of owner financials in non-third party cases

t-kwasnik:
## Bug Fix
In non-third party cases the owner tax and discount percents were not saved to the database resulting in inaccurate after incentive cost calculations in the web UI


sakshi-mishra (approved):
The changes look good, thanks @t-kwasnik !

PR #150, opened 2020-11-19
Chpbeta add cutstrategy solver setting

zolanaj:

This is a proposed model enhancement intended to speed up the tool.  Specifically, we are evaluating a solver setting for Xpress that in preliminary tests by CSM has demonstrated improved optimality gaps for difficult instances within 5 minutes.

No open issue is addressed by this PR.

The new solver setting will change the time required to obtain near-optimal solutions.  This is likely to exhibit improvement for more complex problems (e.g., CHP with Hot TES) but at the cost of slightly longer times for problems with fewer or non-CHP technologies.

No breaking changes

* **Other information**:
@Bill-Becker @t-kwasnik if one of you are able to run the benchmarking on this and the results are available by run to obtain trends for, say, performance change as a function of the collection of technologies available or number utility tiers, that would help with a more refined policy if improvement isn't apparent at the first pass.

### DIFF
--- a/julia_envs/Xpress/precompile.jl
+++ b/julia_envs/Xpress/precompile.jl
@@ -19,5 +19,5 @@ Pkg.build("Xpress")
 include(joinpath(ENV["PROJECT_PATH"], "reo", "src", "reopt_xpress_model.jl"))
 include(joinpath(ENV["PROJECT_PATH"], "reo", "src", "reopt.jl"))
 
-model = reopt_model(300, 0.001)
+model = reopt_model(300, 0.001, 3)
 results = reopt(model, mi)
--- a/reo/src/reopt_decomposed.jl
+++ b/reo/src/reopt_decomposed.jl
@@ -883,7 +883,7 @@ end
 
 function add_decomp_model(m, p::Parameter, model_type::String, mth::Int64)
 	if m[:solver_name] == "Xpress"
-		sub_model = direct_model(Xpress.Optimizer(MAXTIME=-p.DecompTimeOut, MIPRELSTOP=p.DecompOptTol, OUTPUTLOG = 0))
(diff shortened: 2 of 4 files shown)


Bill-Becker:
@zolanaj Thanks for the PR! I think this new solver parameter has to be put into here for the julia image to precompile, right (I typically don't see this issue until I try to build the branch on one of our servers because I don't precompile locally)?
https://github.com/NREL/REopt_Lite_API/blob/4f0d39ae9efa18400f30e25722ae64c65cd2eda9/julia_envs/Xpress/precompile.jl#L22

zolanaj:
@Bill-Becker Thanks for catching; I hadn't included it in the precompile because the solver setting was hard-coded.  I've adjusted it in commit d969d924fa5a39d1991f5336d100ff9aedf0a0ce, but am not making it a custom option yet.  My guess is that this is best kept under the hood and the technology specifications and steplength will determine the appropriate strategy rather than user input.

Bill-Becker:
@zolanaj I ran the same 16 cases that I ran on the chp-merge-decomp branch so I could quickly compare with the latest chpbeta. Four out of the 16 cases timed out with the new cut strategy setting, and there is an overall slowdown for the cases that solved in both.

reopt_seconds   (s) | Avg Techs| Avg Techs Subset* | Avg BAU
-- | -- | -- | --
chpbeta repeat | 87 | 75 | 30
chpbeta_cutstrat | 85 | 85| 31 

*This is the subset of cases which solved for both chpbeta and chpbeta_cutstrat. Four cases timed out on chpbeta_cutstrat which solved on chpbeta (increase the average time on chpbeta because the average for those cases was 120 sec)

zolanaj:
@Bill-Becker thanks for running these. I've asked CSM to run all the different values for this solver setting to see if we can identify a policy for setting this parameter, since this doesn't appear to offer a benefit for all cases (and on average does the opposite).  I'll follow up as we find out more.

PR #153, opened 2020-11-30
Chpbeta unavailability

Bill-Becker:

Feature

No unavailability time to account for CHP downtime due to scheduled and "unscheduled" maintenance where CHP is not able to produce electric or thermal power during specified timesteps.

1. Add an input option `CHP.chp_unavailability_hourly` which has a default profile of unavailability.
a. The unavailability schedule adjusts with the analysis start year to ensure that the weekday types of the default unavailability profile are preserved.
b. Unavailability intervals are **ignored** if it intersects with a resilience outage period
2. The `chp_defaults` endpoint was modified to add the default `chp_unavailability_hourly` profile to the response if the request includes the year of the analysis. This is for the UI to know and have the latest updated CHP unavailability profiles, which will differ based on `CHP.prime_mover`.

No, but the results will change for an identical post that was made before these changes because there is now a default non-zero unavailability profile which hurts CHP economics.

* **Other information**:


Bill-Becker:
@NLaws @danolis This PR introduces unavailability time intervals for CHP in the API. The default CHP unavailability profile is applied in `validators.py` because it depends on the `year` and `CHP.prime_mover`. We don't yet have the actual profiles we will use by `prime_mover`, so a placeholder profile is currently used.

reo/validators.py
```diff
@@ -1053,6 +1085,14 @@ def check_special_cases(self, object_name_path, template_values=None, real_value
                             if v is None:
                                 self.input_data_errors.append('CHP is missing a value for the {} parameter'.format(k))
 
+                        if real_values.get("chp_unavailability_hourly") is None:
+                            self.input_data_errors.append('Must input an 8760 profile for CHP unavailability')
```
NLaws, line 1089: When does this case pop-up? (I thought that we are providing default values for `chp_unavailability_hourly`?)

reo/views.py
```diff
@@ -590,6 +597,27 @@ def chp_defaults(request):
...
+                # TODO put in "prime_mover" instead of hard-coded "recip_engine" for path (after adding other prime_mover unavailability periods)
+                chp_unavailability_path = os.path.join('input_files', 'CHP', 'recip_engine_unavailability_periods.csv')
+                chp_unavailability_periods = pd.read_csv(chp_unavailability_path)
+                if calendar.isleap(year):  # Remove last day of the year if leap year
+                    end_date = "12/31/"+str(year)
+                else:
+                    end_date = "1/1/"+str(year+1)
+                dt_profile = pd.date_range(start='1/1/'+str(year), end=end_date, freq="1H", closed="left")
+                chp_unavailability_hourly = pd.Series(np.zeros(8760), index=dt_profile)
+                for i in range(len(chp_unavailability_periods)):
```
NLaws, line 611: Since this `for` loop is both in views.py and validators.py maybe we could define it in one of the files (or another) and import it where needed? Maybe in utilities.py or something specific to CHP. This would avoid repeated code (DRY or "Don't Repeat Yourself" is a common axiom for code development).

NLaws (changes requested):
Looks good Bill. I have a few comments to be addressed.

danolis:
I've reviewed and the intent is as requested.

danolis (approved):

reo/nested_inputs.py
```diff
@@ -1385,6 +1385,10 @@ def list_of_str(input):
           "max": 1.0,
           "description": "Derate slope as a percent/fraction of rated power per degree F"
         },
+        "chp_unavailability_hourly": {
+          "type": "list_of_float",
+          "description": "CHP unavailability: hourly list (8760) of 1.0 for unavailable and 0.0 for available"
```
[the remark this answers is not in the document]
Bill-Becker: Addressed in [this commit](https://github.com/NREL/REopt_Lite_API/pull/153/commits/06c3f70a3a0a027d963849223c352d2e4981bf5e).

reo/src/techs.py
```diff
@@ -978,9 +981,15 @@ def __init__(self, dfm, run_uuid, existing_boiler_production_type_steam_or_hw, o
...
     @property
     def prod_factor(self):
-        chp_elec_prod_factor = [1.0 for _ in range(8760 * self.time_steps_per_hour)]
+        chp_elec_prod_factor = [1.0 - self.chp_unavailability_hourly[i] for i in range(8760*self.time_steps_per_hour)]
         # Note, we are handling boiler efficiency explicitly so not embedding that into chp thermal prod factor
-        chp_thermal_prod_factor = [1.0 for _ in range(8760 * self.time_steps_per_hour)]
+        chp_thermal_prod_factor = [1.0 - self.chp_unavailability_hourly[i] for i in range(8760*self.time_steps_per_hour)]
+
+        # Ignore unavailability in timestep if it intersects with an outage interval
+        if self.outage_start_hour:
```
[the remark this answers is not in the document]
Bill-Becker: Addressed in [this commit](https://github.com/NREL/REopt_Lite_API/pull/153/commits/a793dd88e3585a7af8f73ece2784a5779955fbdc).

reo/validators.py
Bill-Becker: This is the case where the user does **not** input a `prime_mover`, and in this case, the user must provide the `chp_unavailability_hourly` input as they must for all other cost and performance parameters because we don't have defaults if they don't specify `prime_mover`. We are effectively making `prime_mover` a required input for CHP (if the user doesn't want to input all parameters) but we wanted to allow a completely custom CHP option in which all input parameters are required.

reo/nested_inputs.py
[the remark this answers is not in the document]
NLaws: Looks good @Bill-Becker. Reading the revised description another question dawned on me: can the API handle sub-hourly scenarios with CHP (and this new input)? (i.e. 15 and 30 minute resolution).

reo/views.py
Bill-Becker: I moved this loop/calculation into a new function in `utilities.py`, named it to be more of a general calculation for this purpose, and imported this function for `validators.py` and `views.py`, in [this commit](https://github.com/NREL/REopt_Lite_API/pull/153/commits/9c8a19f0ebdb7a99f9e99fc888568845067dc8cc).

Bill-Becker:
@NLaws Good suggestions and I implemented all of them (see responses to your comment). I "re-requested" your review, but just sending this message in case it's not clear that I addressed your comments.

reo/nested_inputs.py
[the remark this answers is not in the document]
Bill-Becker: @NLaws I believe chp_unavailability_hourly should now work with other time intervals with the updates in [this commit.](https://github.com/NREL/REopt_Lite_API/pull/153/commits/1a16edcd025c7eb7ab73dcdacd404a98a7cbb026) I did not test this yet. Let's punt on making sure the 15-30 minute analysis is functional in general.

reo/validators.py
NLaws: As we discussed offline, currently `prime_mover` is required, and is limited to the `restrict_to` list in nested_inputs.py.  Do you want to keep that behavior and remove the special case handling in the validator, or allow users to not provide the `prime_mover` input? If so you might be able to leverage the `replacement_sets` option in nested_inputs.py.

Bill-Becker: @NLaws thanks for the reminder. I updated the way `CHP.prime_mover` is validated in [this commit to chpbeta](https://github.com/NREL/REopt_Lite_API/commit/8772e8db089c62d89dc80119c0d60df06be975b5) to preserve this case where `CHP.prime_mover` is **not input** while enforcing that all other cost and performance parameters are input.

PR #155, opened 2020-12-02
Test sub-hourly outage, start deprecation of outage values with "hour" in names

NLaws:

Starting the process of using the CHANGELOG.md (who knew that we already had one! Thanks Sakshi!)

### Bug Fixes
- `fuel_used_gal` output for `Generator` was incorrect for scenarios with `time_steps_per_hour != 1`
- `check_common_outputs` was not testing that outputs of type `list` are both `list`s in the expected and calculated outputs

### Changes
- changed `test_flexible_time_steps` to include an outage
- changed check in `validators.py` from "outage_start_hour and outage_end_hour cannot be the same" to "outage_start_hour must be less than outage_end_hour"
- new inputs `outage_start_time_step` and `outage_end_time_step` to replace deprecated `outage_start_hour` and `outage_end_hour`. The latter are used as time step indices in the code, so for sub-hourly problems they do not have hourly units. For now `outage_start_hour` and `outage_end_hour` are kept in place to preserve backwards-compatibility. Also note that the new inputs are not zero-indexed.
- new output `bau_sustained_time_steps` to replace deprecated `sustain_hours` (also not deprecated yet but warning is now in response).

The major issue identified (besides the two bugs above) was that `outage_start_hour` and `outage_end_hour` do not have units of hours when `time_steps_per_hour != 1` - and we limited them to 8759 so outages beyond 8760 time steps could not be modeled with `time_steps_per_hour` of 2 or 4. This problem is fixed with the new inputs. 

Backwards-compatibility is preserved, but we should remove the following once we bump the major version number:
- `outage_start_hour` 
- `outage_end_hour`
- `sustain_hours`

No

* **Other information**:


NLaws:
FYI @Bill-Becker you'll want to merge these bug fixes into chpbeta once Ted approves and we merge to develop.

t-kwasnik:
@NLaws what do you think about renaming outage_start_hour to outage_start_timestep instead of adding new fields to the database?

NLaws:
@t-kwasnik Do you mean change the names in the database but not in the inputs? I think that will work. How will we handle returning user's inputs from the results endpoint if they used the `_hour` values as inputs?

t-kwasnik:
@NLaws Yeah, if you rename the fields in reo/models.py and recreate the migration it should rename the existing db fields. Then, like you have, we keep the validation for either type of input, but only write values in the db to the renamed `_time_step` fields. The results `inputs` section would then only contain the `_time_step` fields, and users can copy and paste this to rerun jobs and still get the same results. The downside would be if anyone needs the `_hour` fields to be in the response inputs section, seems like a pretty rare use case? I think the UI tracks inputs itself so I don't think that workflow would be impacted

NLaws:
@t-kwasnik will https://github.com/NREL/REopt_Lite_API/pull/155/commits/d4f6d9c6ecbe93fe34befd2a6b9d8f1b1834ab63 require a UI change?

NLaws:
@t-kwasnik changing the migrations has somehow affected a recently added field: the `test_summary_and_unlink` is failing with ```null value in column "developer_annual_free_cashflow_series_us_dollars" of relation "reo_financialmodel" violates not-null constraint```.  The `test_summary_and_unlink` appears to be the only test failing since it is the last to run. Can you please look into this?

reo/process_results.py
```diff
@@ -939,9 +939,13 @@ def compute_total_power(self, tech):
         if len(data['outputs']['Scenario']['Site']['PV']) == 1:
             data['outputs']['Scenario']['Site']['PV'] = data['outputs']['Scenario']['Site']['PV'][0]
 
+        
         profiler.profileEnd()
         data['outputs']["Scenario"]["Profile"]["parse_run_outputs_seconds"] = profiler.getDuration()
 
+        #Preserving Backwards Compatability
+        data['inputs']['Scenario']['Site']['LoadProfile']['outage_start_hour'] = data['inputs']['Scenario']['Site']['LoadProfile']['outage_start_time_step']
```
NLaws, line 947: @t-kwasnik outage_start_time_step = outage_start_hour + 1 (because the old inputs are zero-indexed, whereas the new inputs are meant to be more intuitive by being between 1 and 8760 for example)

NLaws:
@t-kwasnik thanks for fixing that! I noticed just now that the develop branch and this branch both have reo migrations with the leading number of 0075, which I believe requires `python manage.py make migrations --merge`? I have run that command locally and am testing now.

t-kwasnik (approved):

PR #156, opened 2020-12-09
Reorganize export and curtail bins, new inputs

NLaws:
### Please check if the PR fulfills these requirements**
- [n/a] Tests for the changes have been added (for bug fixes / features)

Restructure the `SalesTier`s (now `ExportTiers`) in reopt.jl and throughout the model to be more flexible, including four new inputs for each technology:
1. `can_net_meter`
2. `can_wholesale`
3. `can_export_beyond_site_load`
4. `can_curtail`
Generator defaults to False for all four. PV and Wind default to True for all four. (Preserves existing behavior).

- The JuMP model includes three `SalesTier`s in every model with time varying `GridExportRates` available to all technologies
    1. Net Metering
    2. Wholesale
    3. Export beyond annual site load (kWh)
- Net Metering + Wholesale energy is limited to less than or equal to the annual site load
- Any Technology can net meter if `ElecTariff.net_metering_limit_kw > 0` (tier 1)
- Any Technology can participate in a wholesale market, including for energy exported beyond the site load (tiers 2 and 3)
- The third tier is used to track energy curtailed, though users can input a compensation value for the third tier
- Which Technologies can access each of the tiers is hard-coded in `UrdbParse`

- A new decision variable for explicitly tracking curtailed energy
- Refactored `SalesTier`s as `ExportTiers`
- `ExportTiers` can be empty, which results in smaller mathematical models with fewer constraints and decision variables
- Users can choose which Technologies can access which `ExportTiers`
- Set-up of the `ExportTiers` inputs to the JuMP modeled moved from `UrdbParse` to `DataManager`
- removed unused storage export decision variables

### Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
No

### Other
- updated Xpress.jl 0.12.1 -> 0.12.2
- merged with `test_subhourly_outage` in https://github.com/NREL/REopt_Lite_API/pull/155 (which should be merged to develop first)


Bill-Becker:
@NLaws Wow! Major changes, and for the better. I like this direction and strategy in general. My major concern is addressing conflicts with chpbeta. Can you help merging this into chpbeta and aligning the other technologies with this? I can help too in case I did things you would not be sure how to align with this strategy. I made some changes in these areas with CHP exporting, in particular the input for CHP to export was added to ElectricTariff, and not the CHP Tech class. Also I made some changes in chpbeta for how the Generator input for export was handled.

One particular question on <removed unused storage export decision variables> - I know this may have been the case prior, but is storage never (i.e. no user option) to export?

NLaws:
@Bill-Becker yes the storage export decision variables were not used before and were there as a place holder. My understanding is that in general storage is not allowed to export to the grid. And yes I'm happy to help merge with chpbeta.

reo/src/reopt.jl
```diff
@@ -10,8 +10,7 @@ function add_continuous_variables(m, p)
     	dvSystemSizeSegment[p.Tech, p.Subdivision, p.Seg] >= 0   #X^{\sigma s}_{tks}: System size of technology t allocated to segmentation k, segment s [kW]  (NEW)
 		dvGridPurchase[p.PricingTier, p.TimeStep] >= 0   # X^{g}_{uh}: Power from grid dispatched to meet electrical load in demand tier u during time step h [kW]  (NEW)
 	    dvRatedProduction[p.Tech, p.TimeStep] >= 0   #X^{rp}_{th}: Rated production of technology t during time step h [kW]  (NEW)
-	    dvProductionToGrid[p.Tech, p.SalesTiers, p.TimeStep] >= 0  # X^{ptg}_{tuh}: Exports from electrical production to the grid by technology t in demand tier u during time step h [kW]   (NEW)
-	    dvStorageToGrid[p.StorageSalesTiers, p.TimeStep] >= 0  # X^{stg}_{uh}: Exports from electrical storage to the grid in demand tier u during time step h [kW]  (NEW)
+		dvProductionToCurtail[p.Tech, p.TimeStep] >= 0
```
Bill-Becker, line 13: Should we do a similar conditional where this decision variable is only created in the case that some Tech "can_curtail". I.e. if no techs can curtail, avoid creating this dv.

reo/src/reopt.jl
```diff
@@ -540,28 +572,33 @@ function add_storage_grid_constraints(m, p)
...
 	@constraint(m,  AnnualGridSalesLimitCon,
 	 p.TimeStepScaling * ( 
-		sum( m[:dvStorageToGrid][u,ts] for u in p.StorageSalesTiers, ts in p.TimeStepsWithGrid if !(u in p.CurtailmentTiers)) +  sum(m[:dvProductionToGrid][t,u,ts] for u in p.SalesTiers, t in p.TechsBySalesTier[u], ts in p.TimeStepsWithGrid if !(u in p.CurtailmentTiers))) <= p.MaxGridSales[1]
+	  	sum( m[:dvProductionToGrid][t,u,ts] for u in p.ExportTiers, t in p.TechsByExportTier[u], ts in p.TimeStepsWithGrid 
+	  		if !(u in p.ExportTiersBeyondSiteLoad))) <= p.AnnualElecLoadkWh
 	)
 end
 
 
+function add_curtail_constraint(m, p)
```
Bill-Becker, line 594: Maybe this serves to avoid having to conditionally create dvProductionToCurtail (as suggested above) while limiting any solve slowdown.

reo/src/reopt.jl
Bill-Becker: Stringing along these comments as I review, sorry... I now see there is p.TechsCannotCurtail - could we check if this is empty to avoid creating dvProductionToCurtail or would indexing dvProductionToCurtail[{p.Techs} - {p.TechsCannotCurtail}, p.TimeStep] make the dv "Null" if none of the p.Techs can curtail?  Otherwise, I'm fine to keep your "force to zero" constraint for this dV if Tech is in p.TechsCannotCurtail.

Bill-Becker (commented):
@NLaws Looks good, just one comment/question captured between the several comments I left regarding the new dvProductionToCurtail variable in reopt.jl. I'm not necessarily requesting a change, but something to think about if we can avoid creating that variable.

reo/src/reopt.jl
Bill-Becker: Yeah, I think the custom types (e.g. ExportAndCurtailTechs) is a good idea instead of more if/else. However, I also see the advantage of not adding redundant code to capture all the combinations (at the expense of possibly very small amounts of solve time). I don't have a strong opinion on keeping the current modifications versus adding the new Tech types.

Bill-Becker: I now understand better your reference to Julia's method dispatch (I read about it in Julia's documentation after responding). It seems like a pretty smart strategy to start incorporating it, and Julia claims it is one of its most powerful/unique features. I think it would be valuable to introduce that to the model for this use case, and others could learn and re-use that strategy as we develop the model and think about speed-up options. So, without knowing how much, if any, additional and duplicative code this would add, I think it would be beneficial to implement. I'm open to discuss further if need be.

NLaws: @Bill-Becker I just [pushed an outline of the changes needed to use the method dispatch for defining constraints](https://github.com/NREL/REopt_Lite_API/pull/156/commits/d4f0b5a35c7bd603a216a77297da3796194b2b3a). It does not create redundant code (the `...` above are the same code in both the if/else structure or the method dispatch structure).  Once I developed the outline I realized that it will create significant changes to utils.jl and reopt.jl, and so I recommend that we hold off until chpbeta branch is merged into develop branch.

Bill-Becker: @NLaws Sounds good. Thanks for creating the outline. I don't think it's even "must have" before the public release (early/mid January code freeze) since there's a bunch of other stuff we need to make sure we get in there and tested before then.

Bill-Becker (approved):
@NLaws Approving this, and I think the next step is merging this with chpbeta and updating chpbeta to be aligned with it.

PR #157, opened 2020-12-11
SEAC Updates

t-kwasnik:
## New 
Able to parse URDB rates with resolutions finer than hourly
New validation to check the constraint bounds (i.e. that PV min kw is less than max kw)

## Bug Fix
Fix duplicate naming in multiple PV defaults filled in message


reo/validators.py
```diff
@@ -814,6 +815,38 @@ def remove_nones(self, object_name_path, template_values=None, real_values=None,
                         if input_isDict == False:
                             self.input_as_none.append([name, object_name_path[-1] + ' (number {})'.format(number)])
 
+    
+    def check_constraint_ranges(self, object_name_path, template_values=None, real_values=None, number=1, input_isDict=None):
```
NLaws, line 819: Should `input_isDict` be a `bool`?

reo/validators.py
```diff
@@ -1107,6 +1140,16 @@ def check_special_cases(self, object_name_path, template_values=None, real_value
                     electric_tariff['urdb_response'] = rate.urdb_dict
                     self.validate_urdb_response()
 
+            if electric_tariff.get('urdb_response') is not None and len(self.urdb_errors)==0:
+                if len(electric_tariff['urdb_response'].get('energyweekdayschedule',[[]])[0]) > 24:
+                    energy_rate_resolution = int(len(electric_tariff['urdb_response'].get('energyweekdayschedule',[[]])[0]) / 24)
```
NLaws, line 1145: Does the URDB provide schedules that are not hourly? Could you please add a comment or two describing what's happening here? Like `if ...  # then the URDB rate is ...`

reo/validators.py
```diff
@@ -1570,6 +1615,9 @@ def validate_urdb_response(self, number=1):
                     self.urdb_errors += rate_checker.errors
             except:
                 self.urdb_errors.append('Error parsing urdb rate in %s ' % (["Scenario", "Site", "ElectricTariff"]))
+        else:
+            self.urdb_errors.append('Invalid URDB response: %s'.format(str(urdb_response)))
```
NLaws, line 1619: Should this be "Invalid type for `urdb_response`: got {} but must be a `dict`." ?

reo/validators.py
```diff
@@ -1459,15 +1502,17 @@ def fillin_defaults(self, object_name_path, template_values=None, real_values=No
                     if input_isDict or input_isDict is None:
                         self.defaults_inserted.append([template_key, object_name_path])
                     if input_isDict is False:
-                        object_name_path[-1] = object_name_path[-1] + ' (number {})'.format(number)
+                        if not object_name_path[-1].endswith(' (number {})'.format(number)):
```
NLaws, line 1505: Is this just for multiple PV's? I find the `number` name confusing every time I look at the validator. It is an index of the multiple PV's right?

reo/validators.py
NLaws, line 819: What about naming this `check_min_less_than_max` ? And perhaps we should move other checks for min < max to this method? (I know that we have one for outage_start/end_time_step in the validator).

reo/validators.py
```diff
@@ -246,18 +246,18 @@ def validDependencies(self, name):
 
         return valid
 
-    def validCompleteHours(self, schedule_name,expected_counts):
+    def validCompleteHours(self, schedule_name, expected_counts):
```
NLaws, line 249: could you please add a description of the input params to the doc string for this method? It makes it much easier to understand what it is doing.

NLaws (commented):
Some questions and general suggestions to improve code clarity.

reo/validators.py
t-kwasnik: It will effectively always be a bool since this is the only place the function is ever called here: 
https://github.com/NREL/REopt_Lite_API/blob/04310010bfd6829cc37994a6a00b84b64bd55349/reo/validators.py#L656

reo/validators.py
t-kwasnik: It's possible the user specified urdb_label or urdb_utility_name/urdb_rate_name to get the response so I thought it would be better to keep it more generic and just show what the non-compliant content is

reo/validators.py
t-kwasnik: Right now only PV is allowed to be a list or dict, so yes. In the future this is extensible to other inputs like this

t-kwasnik:
@NLaws can you take a look at these recent changes to address your comments?

reo/validators.py
NLaws: Ok then should the default value also be a `bool`?

reo/validators.py
NLaws: @t-kwasnik FYI doc strings have to use triple double quotes to be accessible via built-in `help` https://www.python.org/dev/peps/pep-0257/

NLaws (approved):

reo/validators.py
t-kwasnik: We could make default for input_isDict a bool but I don't see a particular reason for it. With default being None it may help to debug cases where input_isDict isn't set for some reason (we know True/False was explicitly set somewhere)

PR #158, opened 2020-12-15
Chpbeta Absorption Chiller Electric Consumption

Bill-Becker:

Feature

There is no electric consumption for the absorption chiller which is not accurate because absorption chiller still requires heat rejection to cooling towers, which consumes electricity.

Adds an input with a default value for absorption chiller electric consumption (`chiller_elec_cop`), and adds the electric consumption to the load balance constraint in the optimization model.

No, but the results for absorption chiller will change because this new parameter negatively affects absorption chiller.

* **Other information**:
Added an Assert test in test_cold to make sure it's working properly
CHP design document has been updated with this functionality


reo/views.py
```diff
@@ -709,6 +709,8 @@ def chiller_defaults(request):
                         absorp_chiller_opex = defaults_sizes[size - 1][2] + slope_opex * \
                                               (max_cooling_load_tons - defaults_sizes[size - 1][0])
 
+        absorp_chiller_elec_cop = 14.1
```
NLaws, line 712: @Bill-Becker everything looks good. My only concern is that this hard-coded value won't update when someone updates the default value in `nested_inputs.py`. I checked and `nested_input_definitions` is already imported into `views.py`, so I think that you can get the value from the `nested_input_definitions` dict?

NLaws (approved):
One recommended change.

reo/nested_outputs.py
```diff
@@ -947,6 +947,16 @@ def list_of_string(input):
                   "type": float,
                   "description": "Year one chiller thermal production",
                   "units": "TonHr"
+                },
+                "year_one_absorp_chl_electric_consumption_series_kw": {
+                  "type": list_of_float,
+                  "description": "Year one hourly time series of absorption chiller thermal consumption",
```
danolis, line 953: edit "description" to "Year one hourly time series of absorption chiller *electric* consumption"

danolis (changes requested):
Just that one requested change in the code comments

reo/views.py
Bill-Becker: @NLaws Thanks for the review. I linked this value to the `nested_input_definitions` default value for this in [this commit](https://github.com/NREL/REopt_Lite_API/pull/158/commits/ca0df333322943317d39d04da14968c9e00d05af).

reo/nested_outputs.py
Bill-Becker: @danolis thanks for the review. I corrected the description in [this commit](https://github.com/NREL/REopt_Lite_API/pull/158/commits/ca0df333322943317d39d04da14968c9e00d05af).

PR #159, opened 2020-12-16
Update Changelog

t-kwasnik:
Reviewed all PR's since v1.0.0 to put this together (I filtered out less relevant commits i.e. changes to test cases, changes to deployment files). If you want to review your changes to develop feel free, otherwise I think this is comprehensive. 

Please take a look and think about whether this format is going to work for us going forward. Simple changes now can save some one a headache later of trying to reformat this.

When we think this looks good I can tag this commit to release v1.1.0
https://github.com/NREL/REopt_Lite_API/commit/440b13f3fe40b8fa501727db865cfd09343ecc1c

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ Classify the change according to the following categories:
 
 ## develop (unreleased)
 ### Major Updates
+### Minor Updates
 ##### Added
 - `reo`: new inputs **outage_start_time_step** and **outage_end_time_step** to replace deprecated **outage_start_hour** and **outage_end_hour**. The latter are used as time step indices in the code, so for sub-hourly problems they do not have hourly units. For now **outage_start_hour** and **outage_end_hour** are kept in place to preserve backwards-compatibility. Also note that the new inputs are not zero-indexed.
 - `reo`: new output **bau_sustained_time_steps** to replace deprecated **sustain_hours** (also not deprecated yet but warning is now in response).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ Classify the change according to the following categories:
     ##### Removed
     ### Patches
 
-## develop (unreleased) 2020-12-08
+## develop (unreleased)
 ### Major Updates
 ##### Added
 - `reo`: new inputs **outage_start_time_step** and **outage_end_time_step** to replace deprecated **outage_start_hour** and **outage_end_hour**. The latter are used as time step indices in the code, so for sub-hourly problems they do not have hourly units. For now **outage_start_hour** and **outage_end_hour** are kept in place to preserve backwards-compatibility. Also note that the new inputs are not zero-indexed.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,23 +4,187 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Guidelines
+
+Use **bold** markup for field and model names (i.e. **outage_start_time_step**)
+
+Use `code` markup for  REopt Lite-specific file names, classes and endpoints (i.e `reo/validators.py`)
+
+Use _italic_ for code terms (i.e. _list_)
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4,23 +4,187 @@ All notable changes to this project will be documented in this file.
...
+    ##### Added
+    ##### Changed
+    ##### Fixed
+    ##### Deprecated
+    ##### Removed
+    ### Patches
+
+## develop (unreleased)
+### Major Updates
+##### Added
```
Bill-Becker, line 33: @NLaws Are these Major additions, or did the Minor heading just get missed after the Major heading?

Bill-Becker (commented):
@t-kwasnik Do the same #### subheadings (Added, Changed, etc) apply to all of the top ### headings? It kind of looks like the subheadings just apply to the Minor Updates.

CHANGELOG.md
```diff
@@ -4,23 +4,187 @@ All notable changes to this project will be documented in this file.
...
+    
+    ### Major Updates
+    ### Minor Updates
+    ##### Added
+    ##### Changed
+    ##### Fixed
+    ##### Deprecated
+    ##### Removed
+    ### Patches
+
```
kkrah, line 30: looks good to me! One clarification- are "Added, Changed, Fixed, Deprecated, & Removed" only under the "Minor Updates" category or would those ever be used under "Major Updates" and "Patches" too? 

Also, might be helpful to document how to classify something as major, minor, or patch somewhere... not sure if here or somewhere else is the best place but those examples you talked through on the call the other day were helpful for me!

kkrah (approved):

CHANGELOG.md
t-kwasnik: @Bill-Becker I thought it made sense to break up major changes by Added, Changed , Fixed, Deprecated, Removed since these are recommended categories on https://keepachangelog.com/en/1.0.0/. Patches seemed like they would always be minor fixes and upgrades so not sure we wanted to go into as much detail there.

Does that make sense to you? There's no one way to do this, so open to feedback/suggestions

CHANGELOG.md
t-kwasnik: @kkrah take a look at my response to Bill above - I'm not sure we need subheaders for patches, but open to thoughts/suggestions!

We do have some documentation here: https://teams.microsoft.com/l/file/2F639E83-0B4B-48CB-8643-5660704D483E?tenantId=a0f29d7e-28cd-4f54-8442-7885aee7c080&fileType=docx&objectUrl=https%3A%2F%2Fnrel.sharepoint.com%2Fsites%2FREopt%2FShared%20Documents%2FREopt%20OpenSource%2FVersioning%20and%20Release%20Plan.docx&baseUrl=https%3A%2F%2Fnrel.sharepoint.com%2Fsites%2FREopt&serviceName=teams&threadId=19:b1c26b3fe75e4d16ade8e539d580db16@thread.skype&groupId=e4370dd5-3e33-43b4-b753-06b3a269487b

Let me know if the documentation isn't clear or needs more info

Bill-Becker: @t-kwasnik I think there still is a clarification for myself and @kkrah that in the outline/template of the changelog there should be the same set of subheadings under Major as there is currently for Minor, so both Major and Minor should have Added, Changed, Fixed, Deprecated, & Removed.

t-kwasnik: @Bill-Becker @kkrah I wouldn't expect many "major changes" that break backwards compatibility, so I think we should just be able to list the major updates without classifications under that header. 

Maybe instead of classifications at all we add tags like we do for `reo` vs `proforma` for what type of a change it is, then we just append new changes to the top of the list in each Major/Minor/Patches heading?

Bill-Becker: @t-kwasnik @kkrah Ok, that makes sense to me. I could see either way on subheadings versus tagging like `reo`.  In the [above comment](https://github.com/NREL/REopt_Lite_API/pull/159#discussion_r546775278) related to changes that Nick made under the "Major" heading, it seems like he either meant to make those under the Minor heading, or he thought that the same subheadings would apply to Major.

t-kwasnik: @Bill-Becker correct, that was a typo, just fixed it

Bill-Becker (approved):

NLaws (approved):

PR #161, opened 2020-12-30
move avoided outages to outagesimjob

t-kwasnik:
## New

Calculating avoided outage costs from the outagesimjob endpoint

In an upcoming PR, we'll migrate avoided outage costs from the current reo_finanacialmodel table, then remove it from that table. Need to do this first to control the order of adding fields/copying data


Bill-Becker:
@t-kwasnik Can you also merge this into chpbeta? Maybe in general over the next week we should make all changes to chpbeta instead of develop to avoid the two step process of merge to develop then fix merge conflicts when merging develop into chpbeta. chpbeta is up to date with develop as of prior to this PR. I hope to have a PR for develop <-- chpbeta by mid next week.

Bill-Becker (approved):

PR #162, opened 2020-12-31
develop into master 20201231

t-kwasnik:
## new 
- calculate avoided outage costs from `outagesimjob` endpoint
- list_of_int data type
- changed `test_flexible_time_steps` to include an outage	
- changed check in `validators.py` from "outage_start_hour and outage_end_hour cannot be the same" to "outage_start_hour must be less than outage_end_hour"	Prepend change with tag(s) directing where it is in the repository: 
- new inputs `outage_start_time_step` and `outage_end_time_step` to replace deprecated `outage_start_hour` and `outage_end_hour`. The latter are used as time step indices in the code, so for sub-hourly problems they do not have hourly units. For now `outage_start_hour` and `outage_end_hour` are kept in place to preserve backwards-compatibility. Also note that the new inputs are not zero-indexed.	
- new output `bau_sustained_time_steps` to replace deprecated `sustain_hours` (also not deprecated yet but warning is now in response).	
- new `dvProductionToCurtail` in reopt.jl for all techs in all time steps (	`reo`,`proforma`,`resilience_stats`,`*.jl`,`REopt_Lite_API`
    - was previously construed with dvProductionToGrid for the third sales tier, which is meant for selling energy back to the grid beyond the annual load kWh constraint.	
- new inputs for all Techs: `can_net_meter`, `can_wholesale`, `can_export_beyond_site_load`, `can_curtail`

## fixes
- `fuel_used_gal` output for `Generator` was incorrect for scenarios with `time_steps_per_hour != 1`	
- `check_common_outputs` was not testing that outputs of type `list` are both `list`s in the expected and calculated outputs


Bill-Becker (approved):
@t-kwasnik looks like I've reviewed a lot of this stuff already from the PR into develop. Good to see all tests passing.

PR #163, opened 2020-12-31
handle PVWatts download error

t-kwasnik:
## Fixes

Explicitly catch and report on PVWatts download errors


NLaws (approved):
Looks good Ted. I pushed a commit to clarify the logic and remove some redundant code. As soon as the tests pass you can merge.

PR #164, opened 2021-01-04
updated avoided outage cost calc and decprecation message

t-kwasnik:
## Changes
Saving avoided outage costs in the Financial Model to preserve backwards compatibility
Adding deprecation message for this functionality

Note: I moved the deprecation messages to the validator because before we were saving two versions of the warnings message in the db. This way we only save one version of the warnings message because we add the deprecations before initially saving to the database.

### DIFF
--- a/resilience_stats/api.py
+++ b/resilience_stats/api.py
@@ -40,7 +40,7 @@ from tastypie.serializers import Serializer
 from tastypie.validation import Validation
 
 from reo.exceptions import SaveToDatabase
-from reo.models import ScenarioModel, FinancialModel, MessageModel
+from reo.models import ScenarioModel, FinancialModel
 from resilience_stats.models import ResilienceModel
 from resilience_stats.validators import validate_run_uuid
 from resilience_stats.views import run_outage_sim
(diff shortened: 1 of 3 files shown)


t-kwasnik:
@NLaws I need to merge develop to master today to fix that bug Emma found, if this looks good and you can approve this soon I can get this deployed at the same time

resilience_stats/api.py
```diff
@@ -131,6 +131,26 @@ def run_outage_sim_task(scenariomodel_id, run_uuid, bau):
 
     try:
         ResilienceModel.objects.filter(scenariomodel_id=scenariomodel_id).update(**results)
+        results = {'avoided_outage_costs_us_dollars': results['avoided_outage_costs_us_dollars']}
+        FinancialModel.objects.filter(run_uuid=run_uuid).update(**results)
+        deprecation_message = ("'Saving avoided outage costs within the Financial model, accessed from the /results endpoint,"
```
NLaws, line 136: @t-kwasnik I don't think that this message will make sense from a user's perspective since they know nothing of the database back-end. What do you think of putting a different deprecation message about the `avoided_outage_costs_us_dollars` moving from the results to the resilience_stats in the validator? (To replace this deprecation message.)

t-kwasnik:
@NLaws how does this new message look?

NLaws (approved):

PR #166, opened 2021-01-05
fix outage_start/_end_hour inputs

NLaws:
- [na] Tests for the changes have been added (for bug fixes / features)
- [na] Docs have been added / updated (for bug fixes / features)

Bug fix

reo/api.py saves the input_dict to the database with all values in reo/nested_inputs.py. However, `outage_start/end_hour` are no longer in the database but we have kept them in accepted inputs for backwards compatibility. So when api.py tries to save the inputs to the database it fails due to the missing fields for `outage_start/end_hour`.

The `outage_start/end_hour` fields are now deleted from the input_dict in the validator before api.py saves to the database.

No

* **Other information**:

### DIFF
--- a/reo/models.py
+++ b/reo/models.py
@@ -833,5 +833,14 @@ class ModelManager(object):
 
         if resp['inputs']['Scenario']['Site']['LoadProfile'].get('doe_reference_name') == '':
             del resp['inputs']['Scenario']['Site']['LoadProfile']['doe_reference_name']
+        
+        #Preserving Backwards Compatability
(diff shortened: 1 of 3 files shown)


t-kwasnik:
@NLaws this code here is no longer functional

https://github.com/NREL/REopt_Lite_API/blob/11720539799c62c1173b010eb8a16d0a45cc5c12/reo/process_results.py#L988

I moved to the logic to the ModelManager make_response function so that outage start/end hour are not always null in the response. When tests pass this is good to merge

t-kwasnik (approved):

PR #169, opened 2021-01-06
catch bau error in resilience_stats

t-kwasnik:
## Bug Fix

Addresses error in BAU case in resilience_stats endpoint when trying to parse a non-existent BAU avoided outage cost key

### DIFF
--- a/resilience_stats/views.py
+++ b/resilience_stats/views.py
@@ -303,12 +303,13 @@ def run_outage_sim(run_uuid, with_tech=True, bau=False):
         present_worth_factor = annuity(financial.analysis_years, financial.escalation_pct,
                                        financial.offtaker_discount_pct)
     
-    avoided_outage_costs_us_dollars = round( results['resilience_hours_avg'] * \
+    if 'resilience_hours_avg' in results.keys():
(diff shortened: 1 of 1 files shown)


resilience_stats/views.py
```diff
@@ -303,12 +303,13 @@ def run_outage_sim(run_uuid, with_tech=True, bau=False):
         present_worth_factor = annuity(financial.analysis_years, financial.escalation_pct,
                                        financial.offtaker_discount_pct)
     
-    avoided_outage_costs_us_dollars = round( results['resilience_hours_avg'] * \
+    if 'resilience_hours_avg' in results.keys():
+        avoided_outage_costs_us_dollars = round( results['resilience_hours_avg'] * \
```
NLaws, line 307: @t-kwasnik this change makes sense, but I wonder if we should also have `avoided_outage_costs_us_bau_dollars`, which would use the `resilience_hours_avg_bau` value created on line 294?

t-kwasnik: @NLaws I actually had that in there but took it out because we haven't been calculating it already. I just wanted to focus on fixing this bug so the production app will work and then see later if we want to add new fields

NLaws (approved):

Issue #66, opened 2020-06-25
REopt objective needs to be set to keep storage SOC high


NLaws:
Currently the `obj` value is hardcoded to 1, which minimizes LCC. However the production API uses `obj=2`, which also maximizes the average storage state-of-charge so that storage dispatches "look" more reasonable (otherwise the model will typically choose to charge the battery just-in-time for peak shaving, etc.).

Changing `obj` from 1 to 2 will require updating expected values for `test_resilience_stats.py`.

zolanaj:
We might want to consider a structural constraint, say, that the average state-of-charge must be 50% of capacity or greater.  Incentivizing storage levels via the objective function might require parameter tuning, i.e., the incentive on a per-kWh basis might yield results for some cases but not others, and bad incentive values might yield unnecessary (or rather, sub-optimal) battery purchases in some cases.  If this seems reasonable, I can implement this.

NLaws:
@zolanaj We have implemented this in the past with (in pseudo code):
```
dvMeanSOC = sum(ts in TimeStep) dvStoredEnergy(ts) / TimeStepCount
...
objective = min(LCC - dvMeanSOC)
```
which has worked well and not impacted the storage system purchase decision variables. 
Dividing by `TimeStepCount` keeps the value small compared to the LCC. If you divide by `dvStorageKWH` then the problem becomes non linear but maybe you have something else in mind.

Yes please implement if you have time!

zolanaj:
The constraint would look something like the following (in pseudocode that probably breaks mosel and julia): 

@constraint( sum(ts in TimeStep) dvStoredEnergy(ts) / TimeStepCount == 0.5 * dvStorageSizeKWH )

So, it would be a linear constraint (and we'd just substitute out dvMeanSOC from the above).  My understanding is that when using the battery is used for load balancing purposes it can be beneficial to keep it at 50% so there's flexibility for both charging and discharging the battery.  Dispatch might be markedly different for load balancing, peak shaving, and utilizing renewables when resources are greater than load, but the battery system should be able to handle this constraint in every case.

I think we'll want to implememt analogous constraints for the thermal storage system.  Would you agree?

NLaws:
I worry that adding this equality constraint will make the problem more difficult to solve than the old method, but we can try it and test the impact on scenarios with storage in the solution (I think that `test_complex_incentives` has PV and storage in the solution).

zolanaj:
[...]

PR #173, opened 2021-01-08
Address Issue #66: Add optional battery SOC incentive to objective function

zolanaj:

Feature to fix issue https://github.com/NREL/REopt_Lite_API/issues/66

Currently there is an option to add an incentive to the objective function equal to the average SOC of the battery in kWh.  This option, titled "Obj" in the JuMP code, is hard-coded to different values in the production (incentive) and development (no incentive) environments.  When the incentive is included, SOC is divided by 8760 to calculate it, and this is not removed when calculating LCC.

- An optional feature, via the input "add_soc_incentive", has been added to the Scenario inputs as a boolean.  It is fed to the model as part of the parameter object. 
- When included, the sum of hourly SOC is divided by p.TimeStepCount to keep the incentive consistent between non-hourly and hourly problem instances.
- This incentive is removed from the LCC calculation when the incentive is active.

No breaking changes

* **Other information**:
- The default is currently set to "true", i.e., the incentive is added.


reo/src/reopt.jl
```diff
@@ -1052,7 +1051,12 @@ function reopt_run(m, p::Parameter)
     #############  		Outputs    									 #############
     ##############################################################################
 	try
-		results["lcc"] = round(JuMP.objective_value(m)+ 0.0001*value(m[:MinChargeAdder]))
+		if (p.AddSOCIncentive == 1)
```
NLaws, line 1054: @zolanaj I don't think that we need this if/else block: can we use 
``` m[:REcosts] + 0.0001 * value(m[:MinChargeAdder]```
in both cases?

File: reo/nested_inputs.py [hunk 2]
```diff
@@ -160,6 +160,11 @@ def list_of_dict(input):
       "default": 120,
       "description": "The number of seconds allowed before the decomposition subproblem optimization times out"
     },
+    "add_soc_incentive": {
+      "type": "bool",
+      "default": True,
+      "description": "Adds a small incentive for battery state of charge (e.g., to dispatch for resilience) if true, and removes the incentive (e.g., to dispatch for peak-shaving) otherwise"
```
NLaws, line 166: @zolanaj I think that the second example of "to dispatch for peak-shaving" is misleading since the objective is always to minimize cost, which includes peak-shaving if there are demand charges.  Perhaps update the description to: "If True then a small incentive to keep the battery state-of-charge high is added to the objective of the optimization." ?

zolanaj: Thanks for your feedback on this; the description has been updated in commit 5457c1f9dca8bb1bc66317e683583a723f981755

reo/src/reopt.jl
zolanaj: I thought we wanted to remove the incentive when calculating LCC, since it isn't a "real" cost but an artifact of the optimization to obtain reasonable dispatch outcomes like removing then adding the <0.0001 * value(m[:MinChargeAdder]) > back in.  I know the correction wasn't there before, but my impression was that it should be.  Am I missing something?

NLaws: Yes the `REcosts` is the LCC without the SOC incentive in all cases.

zolanaj: Ok, then in that case we need to keep the if/else block in; we only add the SOC incentive if the feature is turned on via `p.AddSOCIncentive == 1`, so we should only remove the quantity from the LCC when `p.AddSOCIncentive == 1` as well.

zolanaj: Thanks again for fixing my mental block on this @NLaws!  Commit db25d210c138ddf913afbf1cc1c64d40720e07e8 implements your suggested changes.

zolanaj:
@NLaws When I run test_annual_loads_urls out of the docker by looping through all cities and building types for each test, the tests now pass.  I don't know whether the merge with recent chpbeta updates fixed this, or if there's a difference between how the docker handles tests vs. GitHub actions, but the test seems to pass now.  Please let me know if I should dig into this any further.

Bill-Becker:
@zolanaj That was a recent fix, so it probably got the fixes when you recently merged chpbeta.

NLaws (approved):

PR #176, opened 2021-01-11
update summary/views.py to return more data when exception occurs

NLaws:
add a try/except block to get as many summary fields as possible

### DIFF
--- a/summary/views.py
+++ b/summary/views.py
@@ -249,123 +249,124 @@ def summary(request, user_uuid):
                 if len(message.keys()) > 0:
                     results['messages'][message.get('message_type') or "type"] = message.get('message') or ""
 
-            if results['status'].lower() != "optimal":
-                json_response['scenarios'].append(results)
-                continue
-            
-            site = get_scenario_data(sites, scenario.run_uuid)[0]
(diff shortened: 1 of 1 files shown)


t-kwasnik:
@NLaws right now, those jobs where the load profile inputs were not initially saved will have load = {} in line 254 and be classified as 'Financial', even though we can't say for sure what kinds of jobs those are since the data was not saved. I checked with Andrew and he skips jobs without a focus. Maybe we should catch these cases and have the focus set to null?  we could also make an 'error' focus type, but Andrew would need to add more code to handle that and it seems like more trouble than it's worth to me. 

Also, we haven't run into issues yet, but line 275 `if tariff['urdb_rate_name']:` will throw an error if we ever run into a save issue with that the electric tariff model so we'll want to similarly handle the case where tariff is {} and make the urdb_rate_name null in that case.

NLaws:
@t-kwasnik I'm not sure what you mean by Andrew "skips jobs without a focus"? The behavior of the focus setting in the API has not changed right? Do we need to change its value under some new conditions?

I'll change `if tariff['urdb_rate_name']:` to `if tariff.get('urdb_rate_name'):`

t-kwasnik (approved):

PR #178, opened 2021-01-21
fix cooling flat defaults

t-kwasnik:
## Update
Updating default cooling flat loads to be the average total kwh across all building types in a city


t-kwasnik:
@Bill-Becker updating the cooling flat load defaults

Bill-Becker (approved):

PR #179, opened 2021-01-21
master<-develop 210121 chpbeta

Bill-Becker:
Merge develop into master after chpbeta was merged into develop.


reo/src/data_manager.py
```diff
@@ -105,6 +105,7 @@ def add_load(self, load):
         self.LoadProfile["resilience_check_flag"] = load.resilience_check_flag
         self.LoadProfile["bau_sustained_time_steps"] = load.bau_sustained_time_steps
         self.LoadProfile["annual_kwh"] = load.annual_kwh
+        self.LoadProfile["loads_kw"] = load.load_list
```
NLaws, line 108: @Bill-Becker this value is redundant with the one on line 103 above. Can the location that we are using "loads_kw" be switched to use "year_one_electric_load_series_kw"?

Bill-Becker: I implemented this because @t-kwasnik told me it's needed for an API response's `inputs` to be used as a POST for another API run. Ted - if I misinterpreted or didn't give the full need, please let us know.

reopt_api/staging_settings.py
```diff
@@ -112,7 +112,7 @@
     'default':{
         'ENGINE': 'django.db.backends.postgresql_psycopg2',
         'HOST': staging_database_host,
-        'NAME': staging_database_name,
+        'NAME': 'reopt_staging',
```
NLaws, line 115: @Bill-Becker just a reminder that we need to switch this staging setting back to use the keys.py value at some point

Bill-Becker: I changed it back to the keys.py value.

NLaws (approved):

Bill-Becker:
@NLaws @t-kwasnik should we deploy master to a wiped db on the staging server (e.g. reopt_chp_staging_internal, after dropping/wiping) **before** we merge this, and then try to merge and deploy to the same? Or are there other plans for trying to emulate this merge on production?

t-kwasnik:
@Bill-Becker I don't see much value in that, it would be the same as doing that locally. I'd say we deploy to production at a time when we're all available to quickly address any issues that might come up.

PR #181, opened 2021-01-22
New om replacement present costs output, after-tax

NLaws:


reo/process_results.py
```diff
@@ -695,13 +198,21 @@ def __init__(self, results_dict, results_dict_bau, dm, inputs):
...
-            replacement_costs += self.inputs["Storage"]["replace_cost_us_dollars_per_kwh"] * \
+            future_cost_storage = self.inputs["Storage"]["replace_cost_us_dollars_per_kwh"] * \
                                  self.nested_outputs["Scenario"]["Site"]["Storage"]["size_kwh"]
-            return round(replacement_costs, 2)
+            future_cost = future_cost_inverter + future_cost_storage
+
+            tax_rate = self.inputs["Financial"]["owner_tax_pct"]
+            discount_rate = self.inputs["Financial"]["owner_discount_pct"]
+            present_cost = 0
+            present_cost += future_cost_inverter * (1 - tax_rate) / ((1 + discount_rate) **
```
Bill-Becker, line 211: @NLaws I'm generally fine leaving as is, but we could make a function (e.g. `future_to_present_value`) for this in `utilities.py`. We also use this equation inside of a `utilities.py` function [here](https://github.com/NREL/REopt_Lite_API/blob/a4683d63491d4120bd24b85e79378d18b2b1f828/reo/utilities.py#L146). Seems like it could be a useful function similar to the `annuity` function already in `utilities.py` for "future" use (pun intended).

Bill-Becker (approved):
@NLaws looks good, just one optional comment to address.

reo/process_results.py
NLaws: Good idea. I'll leave it as-is for now: I personally like the transparency here of how the present cost is determined.

PR #182, opened 2021-01-27
Add Cplex option

NLaws:


NLaws:
@zolanaj I thought maybe you have Cplex such that you can test this new capability?  At the very least I think that you are best positioned to confirm that I got the right Cplex setting for the gap tolerance.

zolanaj:
@NLaws Thanks or putting this together! Your CPLEX parameters are correct.  Should we add a docker file and instructions for CPLEX to the Wiki as well?  

My CPLEX license is for windows, which will make testing this tricky.  I can also work on getting this installed locally on my Windows machine later this week as well if that helps.

NLaws:
@zolanaj Oh yes I can put the Dockerfile for CPLEX together (note that it will require a cplex-linux.bin file that we cannot provide).

NLaws:
@zolanaj I have to put down adding a cplex docker work flow - already spent too much time on it. See the last commit message for a summary of where I left it off.

NLaws:
this is obsolete with new Julia micro service used to address PyJulia threading issues

PR #186, opened 2021-01-29
Coincident peak loads

hdunham:

Coincident peak charge feature

Coincident peak charge not modeled

Coincident peak rates and expected time steps can be specified. There can be a single rate and list of time steps. Or there can be multiple CP periods in a year with different rates, and then a set of time steps is specified for each rate. Peak demand occurring during each set of CP time steps is charged at the corresponding CP rate. 

No

* **Other information**:


summary/views.py
```diff
@@ -310,58 +354,142 @@ def get_scenario_data(data, run_uuid):
...
+                        results['resilience_hours_avg'] = 'not evaluated'
+                    
+                    results['year_one_savings_us_dollars'] = year_one_costs_bau - year_one_costs
 
                     # PV Size
                     if pv is not None:
                         if pv['max_kw'] > 0:
                             results['pv_kw'] = pv.get('size_kw')
+                        else:
+                            results['pv_kw'] = 'not evaluated'
```
t-kwasnik, line 396: @hdunham is there a specific reason you added the 'not evaluated'? I'm going to check with Andrew that the current UI can handle that

t-kwasnik: @hdunham checked with Andrew and these should be null

t-kwasnik (approved):

PR #187, opened 2021-01-29
develop > master 20210129

t-kwasnik:
##New
Coincident peak rates and expected time steps can be specified. There can be a single rate and list of time steps. Or there can be multiple CP periods in a year with different rates, and then a set of time steps is specified for each rate. Peak demand occurring during each set of CP time steps is charged at the corresponding CP rate.

New inputs: 
ElectricTariff > coincident_peak_load_active_timesteps
ElectricTariff > coincident_peak_load_charge_us_dollars_per_kw

New outputs: 

ElectricTariff > year_one_coincident_peak_cost_us_dollars
ElectricTariff > year_one_coincident_peak_cost_bau_us_dollars
ElectricTariff > total_coincident_peak_cost_us_dollars
ElectricTariff > total_coincident_peak_cost_bau_us_dollars


reo/validators.py
```diff
@@ -1390,6 +1399,25 @@ def check_special_cases(self, object_name_path, template_values=None, real_value
...
+            if self.isValid:
+                if electric_tariff.get('coincident_peak_load_active_timesteps') is not None:
+                    for series in electric_tariff.get('coincident_peak_load_active_timesteps'):
+                        self.validate_timestep_series(series, 
+                        "ElectricTariff", 'coincident_peak_load_active_timesteps', 
+                        ts_per_hour, number=number, input_isDict=input_isDict)
+                    if len(electric_tariff.get('coincident_peak_load_active_timesteps')) != len(electric_tariff.get('coincident_peak_load_charge_us_dollars_per_kw')):
+                        self.input_data_errors.append(( "The number of rates in coincident_peak_load_charge_us_dollars_per_kw must"
+                                                        " match the number of timestep sets in coincident_peak_load_active_timesteps"))
+                    if self.isValid:
```
dylancutler, line 1412: is this second self.isValid necessary?

summary/views.py
```diff
@@ -343,25 +408,76 @@ def get_scenario_data(data, run_uuid):
...
-                    # ColdTES Size
-                    if coldtes is not None:
-                        if (coldtes.get('max_gal') or -1) > 0:
-                            results['coldtes_gal'] = coldtes.get('size_gal')
-
-                    # Absoprtion Chiller Size
-                    if absorpchl is not None:
-                        if (absorpchl.get('max_ton') or -1) > 0:
-                            results['absorpchl_ton'] = absorpchl.get('size_ton')
+                        else:
```
dylancutler, line 411: Issue with the "else" here. Please ensure that this merged correctly.

reo/validators.py
t-kwasnik: @dylancutler yes, if both coincident peak inputs are not provided isValid will be false and in that case we don't want to run this additional check.

dylancutler (approved):
This looks addressed now

reo/validators.py
t-kwasnik: @dylancutler yes, we don't want to run that default setting if the data is invalid

PR #191, opened 2021-02-02
hotfix 02022021

t-kwasnik:
## Patches
`reo`: **Wind** `size_class` was not being set
`proforma`: could not handle runs prior to v1.4.0 with no CHP database entries
`resilience_stats`: could not handle runs prior to v1.4.0 with no CHP database entries


t-kwasnik:
@Bill-Becker A few more hot fixes for errors that came up in Rollbar for missing CHP entries in the DB (entries prior to v1.4.0)

Bill-Becker (approved):

PR #192, opened 2021-02-03
develop > master 20210203

t-kwasnik:
### Patches
- `reo`: Fix **Wind** `size_class` was not being set
- `proforma`: Fix could not handle runs prior to v1.4.0 with no CHP database entries
- `resilience_stats`: Fix could not handle runs prior to v1.4.0 with no CHP database entries
- `resilience_stats`: `outage_simulator` returns 100% survivability when chp_kw >= critical_loads_kw

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ Classify the change according to the following categories:
     
 ## v1.4.1 - 2021-01-29
 ### Patch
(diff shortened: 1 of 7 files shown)


t-kwasnik:
@Bill-Becker this is to deploy those v1.4.0 fixes you approved, plus one more from Nick

Bill-Becker (approved):

PR #193, opened 2021-02-05
Nginx docker-compose, PVWatts bug fix

NLaws:
### Patches
- `reo`: Fix **PVWatts** being called when user provides `PV.prod_factor_series_kw`
- `reopt_api`: new `docker-compose.nginx.yml` for standing up the API on a server with remote access (for example if one wants to host the API on a cloud service)
- `reopt_api`: update `Dockerfile.xpress` to use `nlaws/pyjul:1.5.2` base image (was using Julia 1.3)

### DIFF
--- a/reo/src/pvwatts.py
+++ b/reo/src/pvwatts.py
@@ -48,7 +48,8 @@ def check_pvwatts_response_data(resp):
         message = ("Error parsing PVWatts response. "
                    "Status code: {}. Text: {}".format(resp.status_code, resp.text))
         raise_pvwatts_exception(message)
-    return data
+    else:
+        return data
 
 
 def raise_pvwatts_exception(message):
@@ -85,7 +86,29 @@ class PVWatts:
                  offline=False,
                  verify=True,
                  **kwargs):
-
+        """
+        WARNING: the __init__ method calls PVWatts API
+        :param url_base: str
+        :param key: str
+        :param azimuth: float
+        :param system_capacity: float
+        :param losses: float
+        :param array_type: int
+        :param module_type: int
+        :param timeframe: str
+        :param gcr: float
+        :param dc_ac_ratio: float
+        :param inv_eff: float
+        :param radius: float
+        :param dataset: str
+        :param latitude: float
+        :param longitude: float
+        :param tilt: float
+        :param time_steps_per_hour: int
+        :param offline: bool
+        :param verify: bool
+        :param kwargs: dict
+        """
         self.url_base = url_base
         self.key = key
         self.azimuth = azimuth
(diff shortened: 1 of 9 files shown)


reo/src/techs.py
```diff
@@ -146,7 +148,10 @@ def __init__(self, dfm, degradation_pct, time_steps_per_hour=1, acres_per_kw=6e-
             else:  # All other tilts come from lookup table included in the array_type_to_tilt_angle dictionary above
                 self.tilt = PV.array_type_to_tilt_angle[kwargs.get('array_type')]
 
-        self.pvwatts = PVWatts(time_steps_per_hour=self.time_steps_per_hour, azimuth=self.azimuth, tilt=self.tilt, **self.kwargs)
+        if self.prod_factor_series_kw is not None:  # then don't call PVWatts
+            self.station = (kwargs.get("latitude", 0), kwargs.get("longitude", 0), 0)
```
t-kwasnik, line 152: Do we need to set station coords? Lat and long are required so we shouldn't get (0,0), but that is is a real point and may be confusing if it comes back from the API. Is there a problem with (None, None)?

reo/src/pvwatts.py
```diff
@@ -48,7 +48,8 @@ def check_pvwatts_response_data(resp):
         message = ("Error parsing PVWatts response. "
                    "Status code: {}. Text: {}".format(resp.status_code, resp.text))
         raise_pvwatts_exception(message)
-    return data
+    else:
```
t-kwasnik, line 51: Just curious, why was this change needed? If an exception is encountered we'll exit out of this workflow through the raise funciton

t-kwasnik (commented):
@NLaws left two quick questions, can approve soon after you get to those

reo/src/pvwatts.py
NLaws: Two reasons: 
1. clarity: the `else` statement says if no exception occurs then do this
2. robustness: if the `except` block is ever changed to something that does not raise an exception then the code would fail (since `data` would not be defined due to the scoping of the try/except block)

reo/src/techs.py
NLaws: I was trying to preserve old behavior, but we don't need to set them. (Old behavior was to call PVWatts to get station lat/lon even when user provider their own prod_factor).

NLaws:
the Julia upgrades might fix https://github.com/NREL/REopt_Lite_API/issues/183 (I have run 3,000 jobs on this branch and seen no lost jobs), which include:
- Julia 1.5.3 available in nlaws/pyjul images (we will have to manually upgrade on RHEL servers)
- PyCall.jl 1.91.4 >> 1.92.2

reo/src/techs.py
t-kwasnik: @NLaws thanks - I'll approve this and check with Andrew about the 0,0 issue and decide how to handle that after that

t-kwasnik (approved):

PR #195, opened 2021-02-16
Third party O&M cost output

NLaws:
new output requested for the UI


NLaws:
@danolis  I added the new output to the API because there is no straightforward way to calculate the third party factor in the UI

danolis (approved):
This looks good to me!

PR #196, opened 2021-02-17
master<-develop 210217

NLaws:
- `reo`: new output `Financial.developer_om_and_replacement_present_cost_after_tax_us_dollars`
- `reo`: Fix **PVWatts** being called when user provides `PV.prod_factor_series_kw`
- `reopt_api`: new `docker-compose.nginx.yml` for standing up the API on a server with remote access (for example if one wants to host the API on a cloud service)
- `reopt_api`: update `Dockerfile.xpress` to use `nlaws/pyjul:1.5.3` base image (was using Julia 1.3)
- `reopt_api`: update `julia_envs/Xpress` PyCall from 1.91.4 to 1.92.2


CHANGELOG.md
```diff
@@ -28,6 +28,14 @@ Classify the change according to the following categories:
     ##### Removed
     ### Patches
     
+## dev - 2021-02-16
```
t-kwasnik, line 31: @NLaws since this is going into master, at this point we can tag the release number. I think this make sense because if we wait till after the PR goes through to update the number here develop and master will show up as being out-of-sync

t-kwasnik (approved):
@NLaws approved! Just think about addressing the version number before merging

PR #197, opened 2021-02-19
Hot fix 02/19/21

t-kwasnik:
## Patches
- `reo`: Fix divide by 0 error in results processing
- `reo`: Handle floats as URBD periods
- `reo`: Fix `list_of_float` only types

### DIFF
--- a/reo/process_results.py
+++ b/reo/process_results.py
@@ -789,12 +789,12 @@ def process_results(self, dfm_list, data, meta, saveToDB=True):
                     self.nested_outputs["Scenario"]["Site"]["Financial"][
                         "om_and_replacement_present_cost_after_tax_us_dollars"] / self.third_party_factor
 
-            self.nested_outputs["Scenario"]["Site"]["renewable_electricity_energy_pct"] = \
-                self.nested_outputs["Scenario"]["Site"]["Wind"].get("average_yearly_energy_produced_kwh") or 0
-            for pv in self.nested_outputs["Scenario"]["Site"]["PV"]:
-                self.nested_outputs["Scenario"]["Site"]["renewable_electricity_energy_pct"] += \
-                pv.get("average_yearly_energy_produced_kwh") or 0
             if self.nested_outputs["Scenario"]["Site"]["LoadProfile"]["annual_calculated_kwh"] > 0:
+                self.nested_outputs["Scenario"]["Site"]["renewable_electricity_energy_pct"] = \
+                    self.nested_outputs["Scenario"]["Site"]["Wind"].get("average_yearly_energy_produced_kwh") or 0
+                for pv in self.nested_outputs["Scenario"]["Site"]["PV"]:
+                    self.nested_outputs["Scenario"]["Site"]["renewable_electricity_energy_pct"] += \
+                    pv.get("average_yearly_energy_produced_kwh") or 0
                 self.nested_outputs["Scenario"]["Site"]["renewable_electricity_energy_pct"] = round(
                     self.nested_outputs["Scenario"]["Site"]["renewable_electricity_energy_pct"] /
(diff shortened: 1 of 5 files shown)


reo/process_results.py
```diff
@@ -794,9 +794,10 @@ def get_nested(self):
             for pv in self.nested_outputs["Scenario"]["Site"]["PV"]:
                 self.nested_outputs["Scenario"]["Site"]["renewable_electricity_energy_pct"] += \
                 pv.get("average_yearly_energy_produced_kwh") or 0
-            self.nested_outputs["Scenario"]["Site"]["renewable_electricity_energy_pct"] = round(
-                self.nested_outputs["Scenario"]["Site"]["renewable_electricity_energy_pct"] /
-                self.nested_outputs["Scenario"]["Site"]["LoadProfile"]["annual_calculated_kwh"], 4)
+            if self.nested_outputs["Scenario"]["Site"]["LoadProfile"]["annual_calculated_kwh"] > 0:
```
NLaws, line 797: @t-kwasnik there might be some other issue at hand here: when is `annual_calculated_kwh` ever not `> 0`? Seems like it should always be some positive number? Or can it be negative if existing PV is huge, which would imply that  `renewable_electricity_energy_pct` is greater than 100% (making `annual_calculated_kwh` < 0). So if `annual_calculated_kwh` is < 0 then we need to determine `renewable_electricity_energy_pct another way?

t-kwasnik: @NLaws I personally don't see a need to allow load profiles to be all 0's, but do remember Dan had been running some 0 load tests within the past few months and wanted this functionality. 

Also, by the time we have populated `annual_calculated_kwh` we have put back in the existing PV if the initial load provided was net, so we should not see the > 100% issue you mentioned. Also, we check the non-net load saved in `annual_calculated_kwh` for negative numbers within a tolerance and either (a) raise an error if the load goes too negative or (b) reset the slightly negative numbers to 0. 

https://github.com/NREL/REopt_Lite_API/blob/3a161ba4276144654456f5b6da74661426d7a479/reo/scenario.py#L199

In the case where load actually is zero I think it makes sense to keep the result as None (and not NaN or some other value).

Do you see any changes needed now given these clarifications?

Bill-Becker: > @NLaws I personally don't see a need to allow load profiles to be all 0's, but do remember Dan had been running some 0 load tests within the past few months and wanted this functionality.

@t-kwasnik @NLaws Just reiterating that we do want to allow 0's for the **heating** load so that we can evaluate CHP as an electric generator only. We don't have a request to make the **electric** load 0's.

reo/process_results.py
```diff
@@ -789,14 +789,15 @@ def get_nested(self):
...
 
-            self.nested_outputs["Scenario"]["Site"]["renewable_electricity_energy_pct"] = \
-                self.nested_outputs["Scenario"]["Site"]["Wind"].get("average_yearly_energy_produced_kwh") or 0
-            for pv in self.nested_outputs["Scenario"]["Site"]["PV"]:
-                self.nested_outputs["Scenario"]["Site"]["renewable_electricity_energy_pct"] += \
-                pv.get("average_yearly_energy_produced_kwh") or 0
-            self.nested_outputs["Scenario"]["Site"]["renewable_electricity_energy_pct"] = round(
-                self.nested_outputs["Scenario"]["Site"]["renewable_electricity_energy_pct"] /
-                self.nested_outputs["Scenario"]["Site"]["LoadProfile"]["annual_calculated_kwh"], 4)
+            if self.nested_outputs["Scenario"]["Site"]["LoadProfile"]["annual_calculated_kwh"] > 0:
```
t-kwasnik, line 792: @NLaws fixed

NLaws (approved):

PR #200, opened 2021-02-23
Change heating and fuel units to kW/kWh in reopt.jl

Bill-Becker:

1. Change the units-basis for fuel consumption and heating supply/load to kWh in the **reopt.jl model** (did NOT change the units basis for the inputs/outputs).
2. Delete unused/irrelevant inputs for Boiler and ElectricChiller (sizing and cost)

(Units of mmbtu are used for Boiler and CHP fuel consumption/cost and thermal production in the reopt.jl model. Units of gal are used for Generator fuel consumption/cost in the reopt.jl model)

(No new behavior. This change is anticipated to speed up the optimization model because changing fuel and heating to the same units as electricity (kW/kWh) results in less of a numeric difference between the electric and heating production and loads - this is thought to improve numeric stability (per Alex Z. and the Mines team).)

Technically yes, however, the deleted inputs for Boiler and ElectricChiller are basically irrelevant and unused because we use the heating and cooling load to cap the size, and we don't assign a cost for those technologies. Although possible, it is expected that no users have used these inputs up to this point. The Webtool does NOT use these inputs at all. It would be beneficial to remove these confusing inputs now (after less than a month of them being in production) than deal with the hassle of deprecating them later after we get use and questions from them.

* **Other information**:


Bill-Becker:
@zolanaj I wanted to make the switch from MMBtu to kWh for the heating load/supply in reopt.jl. I also changed the fuel consumption and cost to be based on kWh, so that everything is kW/kWh in the model. I'm finding that it's slowing down the CHP runs. I've only been able to test locally because our servers are having trouble with a recent upgrade to Julia with PyJulia/Pycall, but I've been getting pretty consistent results. POSTing the test_chp.CHPTest.test_chp_sizing_monolith_1pct equivalent about 5 times on each on the develop branch and this branch, I'm getting ~130 seconds on develop and ~240 seconds on this branch. So, pretty significant. I wanted to see if you could spot anything obvious which might cause the slowdown. Thanks!

Bill-Becker:
@zolanaj I did some more benchmarking, and I am seeing a slowdown of about 20-25% for both BAU and "Techs" cases. Maybe you could still take a look and see if anything pops out at you.

zolanaj:
@Bill-Becker take a look at the most recent commit and let me know what you think.  The only thing that came to mind was that if we're not considering the size of the electric chiller or boiler then there's no reason to put any constraint on the size or hourly production.  Running it locally reduced the test_chp monolith run and a separate local test run by a few seconds each, and I don't think it will slow anything down, but not sure if it will move the needle on the benchmarking.  Just out of curiosity, is it the julia solve time that's 20-25% higher or is it sourced from other things?

Bill-Becker:
@zolanaj Thanks for taking a look. I did suspect something related to sizing might be an issue. However, we do still need to keep a size/production constraint on Boiler and ElectricChiller because of HotTES/AbsorptionChiller and ColdTES which would otherwise could cause Boiler and ElectricChiller to dispatch well above its rated/expected capacity - this is still done in this branch but only based on the peak/max load for heating (Boiler) and cooling (ElectricChiller). We just find the peak load and use the max_thermal_factor_on_peak_load for those technologies to determine the max size now, without the option to put in a unique max size.

zolanaj:
@Bill-Becker ok, fair enough.  We used to have a "fudge factor" of either 1.25 or 2.0, and it might be good to include that and try again.  My understanding is that these are generally a bit oversized vs. peak load anyhow, is that correct?  It would also prevent infeasibilities from potentially occurring when floating points move from Python to Julia (I can't speak for JuMP directly, but in GAMS you're allowed up to 8 decimals and the rest is truncated, so it's possible).  This might change how Xpress handles the problem (again, not sure about Xpress, but I've seen it before elsewhere).  If that makes sense to you I can implement it pretty quickly and revert the previous commit.

Bill-Becker:
@zolanaj Right, we have the `max_thermal_factor_on_peak_load` input for both Boiler and ElectricChiller that is the multiplier on the peak load for the max size. The default is 1.25 for both.

Bill-Becker:
@NLaws @zolanaj After benchmarking both locally and on the development server (looking at ~10x iterations of several heating-relevant scenarios), it's been determined that there is no slowdown due to these changes. Alex and I carefully looked at the Profile times, and the cause of differences in solve times was determined to be computational resource variance because it affected all aspects of the Profile times consistently (i.e. not just the julia_reopt_optimize_seconds). I also saw a ~25% **speed up** in my most recent benchmarking, but again that should only be attributed to resource variance.

zolanaj (approved):

Bill-Becker:
@NLaws Hi Nick, would you be able to review this by this week?

NLaws (approved):

PR #201, opened 2021-02-24
Hotfix022421

t-kwasnik:
### Patches
- `reo`: In `validators.py` catches case where invalid percent_share entry was used in check special cases function
- `reo`: In `loadprofile.py` catches where 0 lat/long was resolving to False and leading to None for lat and long

### DIFF
--- a/reo/src/load_profile.py
+++ b/reo/src/load_profile.py
@@ -435,9 +435,6 @@ class BuiltInProfile(object):
             # try shapefile lookup
             log.info("Trying city lookup by shapefile.")
             gdf = gpd.read_file('reo/src/data/climate_cities.shp')
-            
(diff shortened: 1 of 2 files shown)


t-kwasnik:
@NLaws 2 more rollbar issues that came in this morning - note that in validators.py it looks like I did a lot of changes but it's mostly just tabbing because of an `if` clause I added

NLaws (approved):
Looks good @t-kwasnik. Just curious: when do we get zero lat or lon values?

t-kwasnik:
@NLaws That point is in the ocean so it doesn't really make sense, but it came from a job a user actually submitted. I suspect there is something wrong with their script, or they are just running a demo case....

PR #204, opened 2021-03-09
Address rollbar issues

t-kwasnik:
## Patches
- `reo`: Catch issue in `process_results.py` where `renewable_electricity_energy_pct` is not explicitly set to _None_
- `reo`:  Catch case where `CHP` `prime_mover` is not set and not all required fields are filled in
- `reo`:  Catch issues with `itc_unit_basis` when the ITC is 100%
- `validators.py`: Fix bug where length of percent_share != length of doe_reference_name even though no percent_share is provided (in `LoadProfileBoilerFuel`)

### DIFF
--- a/reo/validators.py
+++ b/reo/validators.py
@@ -1130,7 +1130,6 @@ class ValidateNestedInput:
 
                         # Do same validation on chp_unavailability periods whether using the default or user-entered
                         self.input_data_errors += ValidateNestedInput.validate_chp_unavailability_periods(year, chp_unavailability_periods)
-
                         self.validate_chp_inputs(updated_set, object_name_path, number)
 
                 # otherwise, check if the user intended to run CHP and supplied sufficient info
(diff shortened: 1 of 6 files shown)


reo/validators.py
```diff
@@ -1162,18 +1161,21 @@ def check_special_cases(self, object_name_path, template_values=None, real_value
                     # check if user intended to run CHP and supplied sufficient pararmeters to run CHP
                     if user_supplied_chp_inputs:
                         required_keys = prime_mover_defaults_all['recip_engine'].keys()
-                        filtered_values = {k: real_values.get(k) for k in required_keys}
+                        if real_values.get('prime_mover') is None:
+                            self.input_data_errors.append('No prime_mover was input so all cost and performance parameters must be input.')
```
t-kwasnik, line 1165: @Bill-Becker does this not cover your comment below?

reo/validators.py
```diff
@@ -1162,18 +1161,21 @@ def check_special_cases(self, object_name_path, template_values=None, real_value
...
+                        filtered_values = {k: real_values.get(k) for k in required_keys if k is not 'prime_mover'}
+                        missing_defaults = []
                         for k,v in filtered_values.items():
                             if v is None:
-                                self.input_data_errors.append('No prime_mover was input so all cost and performance parameters must be input. \
-                                    CHP is missing a value for the {} parameter'.format(k))
-
+                                missing_defaults.append(k)
+                        if len(missing_defaults) > 0:                               
+                            self.input_data_errors.append("CHP is missing a value for the following defaults: " + ', '.join(missing_defaults))
```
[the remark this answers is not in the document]
t-kwasnik: @Bill-Becker if the message above L1164 is not sufficient, what about

CHP is missing a value for the following prime mover specific defaults:

reo/validators.py
t-kwasnik: @Bill-Becker what happens is that it first checks to see that prime_mover is set, if not an error is added to the error array. It doesn't error out though until checking to see what defaults are missing and adding a message about that. The final error message is two sentences: 

"error": "No prime_mover was input so all cost and performance parameters must be input. CHP is missing a value for the following defaults: attr1, attr2, attr3"

Bill-Becker: So if it doesn't find that any of the required inputs are missing, it won't ever error out? It still seems like it will "error" with the "No prime_mover was input so all cost and performance parameters must be input.'" message because that was appended to the self.input_data_errors message.

t-kwasnik: @Bill-Becker nope, appending to the input_data_errors list doesn't immediately trigger an error. That's actually why we first starting looking into this issue. The original problem was that we were hitting L1178 self.validate_chp_inputs when errors had been found and the job was actually in an invalid format. When errors are found they are actually raised in reo/api.py not validators.py

Bill-Becker: @t-kwasnik sorry for my confusion on this, but you said it doesn't **immediately** trigger an error, but are you also saying that if it doesn't find any missing parameters that it will **never** trigger an error? I am going to make a few tweaks on this branch this morning to reduce the number of required inputs for CHP in the case they don't send a prime_mover type - all of this checking will still be needed, but there will be a lesser number of params to pass.

reo/validators.py
```diff
@@ -1162,18 +1161,19 @@ def check_special_cases(self, object_name_path, template_values=None, real_value
...
-                        filtered_values = {k: real_values.get(k) for k in required_keys}
+                        filtered_values = {k: real_values.get(k) for k in required_keys if k not in ['prime_mover', 'size_class']}
+                        missing_defaults = []
                         for k,v in filtered_values.items():
                             if v is None:
-                                self.input_data_errors.append('No prime_mover was input so all cost and performance parameters must be input. \
-                                    CHP is missing a value for the {} parameter'.format(k))
-
+                                missing_defaults.append(k)
+                        if len(missing_defaults) > 0:
```
t-kwasnik, line 1169: @Bill-Becker I believe this addresses what we talked about, please verify

Bill-Becker: Yep, looks right.

Bill-Becker (approved):

PR #215, opened 2021-04-01
Remove decomposition

NLaws:
Just deleting obsolete code.


Bill-Becker (approved):
@NLaws I removed the decomposition related inputs as well. That technically makes it a breaking change, but decomposition has always been broken since we released it anyway, so I don't think there's any issue with removing those.

PR #218, opened 2021-04-13
master<-dev 21.04.13 Julia pods 10->15

NLaws:
based on cpu usage we can scale up 50-75%.  as we scale up towards full utilization of the Rancher hosts we will also be able to better understand how high we can set the user POST rate limits.

also:
- Bump django from 3.1.7 to 3.1.8
- typo fix in Jenkinsfile


NLaws:
@Bill-Becker FYI the `continuous-integration/jenkins/pr-head` fails are unrelated to this PR (debugging testranch right now, but this PR builds OK on rancher).

Bill-Becker (approved):

PR #220, opened 2021-04-22
Chunkedsummary

t-kwasnik:
### Description

This is a request from the UI team for users who have too many jobs to load at once. This way the number of records in each chunk can be parameterized and results pulled back in pieces.


Bill-Becker:
@NLaws I'm finally getting around to reviewing this. I spoke with Andrew J. and he thinks we could resolve the slow /user/summary endpoint which spawned this task by adopting the more efficient database call process that we've been discussing in V2 meetings. Andrew also thought this endpoint is non-intuitive to use for the UI. Ted thought the /summary endpoint was reasonably optimized even though we have multiple calls to the db (I'm interpreting this 2nd-hand, so could be misinterpreting that), and he thought this chunked strategy was a good solution. 

I'm testing this on the chunkedsummary branch on staging, and I don't quite understand the chunk parameter because I'm changing it and it's not giving me back a different amount of runs.

Any thoughts? Maybe we want to have a call to discuss.

NLaws:
@Bill-Becker yes let's chat

Bill-Becker (approved):

Bill-Becker (approved):

PR #222, opened 2021-04-29
master<-dev fix international pvwatts, increase prod julia pods 15->20

NLaws:


reo/src/pvwatts.py
```diff
@@ -131,38 +131,34 @@ def __init__(self,
...
             resp = requests.get(self.url, verify=self.verify)
-
-            if not resp.ok:
-                data = check_pvwatts_response_data(resp)
-                # if not ok b/c outside of the US, try again with larger search radius in "intl" dataset
-                if ("This location appears to be outside the US" in s for s in data.get("warnings", [])):
-                    self.dataset = "intl"
-                    self.radius = self.radius * 2  # bump up search radius, since there aren't many sites
-                    resp = requests.get(self.url, verify=self.verify)
-                    if not resp.ok:  # did not get data for international location either (or other problem)
```
hdunham, line 162: this section that handles the case when  pvwatts response is not ok, why is it deleted?

reo/src/pvwatts.py
```diff
@@ -131,38 +131,34 @@ def __init__(self,
...
+                self.dataset = 'intl'
+                self.radius = self.radius *2
             resp = requests.get(self.url, verify=self.verify)
-
-            if not resp.ok:
-                data = check_pvwatts_response_data(resp)
-                # if not ok b/c outside of the US, try again with larger search radius in "intl" dataset
-                if ("This location appears to be outside the US" in s for s in data.get("warnings", [])):
-                    self.dataset = "intl"
-                    self.radius = self.radius * 2  # bump up search radius, since there aren't many sites
```
hdunham, line 160: would be nice to keep this comment because I did not know why the radius was being doubled until I read it

hdunham (changes requested):

reo/src/pvwatts.py
NLaws: it is no longer needed because we check and fix the dataset argument rather than trying and failing to get values outside the US

hdunham: great

reo/src/pvwatts.py
NLaws: updated the comment

NLaws:
@hdunham The status of this PR shows `changes requested`: is there anything else that I need to address?

hdunham (approved):

PR #228, opened 2021-06-01
BUG FIX urdb_parse TOU demand charge time steps

NLaws:
TOU demand charge time steps were zero indexed, which is incorrect for arrays that are used in reopt_model.jl (should be one indexed)


reo/src/urdb_parse.py
```diff
@@ -639,7 +638,13 @@ def get_hours_in_month(self, month):
...
+        """
+        Get the TOU steps indexed on 1-8760 for hourly simulations
+        :param current_rate: RateData class instance
+        :param month: int
+        :param period: int
+        :return: list of ints
+        """
         step_array = []
         start_step = 1
         start_hour = 1
```
lixiangk1, line 650: delete also?

lixiangk1 (approved):

reo/src/urdb_parse.py
NLaws: is it wrong? (I was trying to document how the function is meant to be used.)

lixiangk1: oh haha, I thought the comment was going to show up more line specific. just the start_hour=1 part

NLaws: gotchya: removed it in https://github.com/NREL/REopt_Lite_API/pull/228/commits/f9e2ba7d5606a6777b329125ef04ae7d9b51f82b

PR #230, opened 2021-06-03
Fix user urls

NLaws:
## Bug fixes
- The `v1/user/<user_uuid>/unlink/<run_uuid>` end point was not working due to unclear path definitions in `summary/urls.py` and `reopt_api/urls.py`
- Once the URLs were fixed, the 204 code forced the `JSONResponse` to be  empty (because that is what is expected for 204). Changed it to 208.
## Changes
-  clarified the `add_user_uuid` view (which has never been used except by `summary/tests/test_summary_and_unlink.py`)
- clarified help text for `Jenkinsfile-undeploy`


NLaws:
@hdunham can you please prioritize this review? It is causing many errors in production

summary/views.py
```diff
@@ -40,65 +39,53 @@
 from summary.models import UserUnlinkedRuns
 
 
-def add_user_uuid(request):
+def add_user_uuid(request, user_uuid, run_uuid):
```
hdunham, line 42: request param no longer needed?

hdunham: wouldn't you also need to change the arguments being passed wherever this function is called? I'm not seeing that.

hdunham (approved):

summary/views.py
NLaws: The `views.py` functions are used by Django as the endpoints and have to implement `request` as the first argument. https://docs.djangoproject.com/en/3.2/topics/http/views/

PR #240, opened 2021-07-02
Offgrid

NLaws:
## TODO
- [ ] keep "on grid" API behavior by setting the new input diesel fuel cost escalation equal to the existing input electricity cost escalation
- [ ] fix test suite
- [ ] fill out PR template below
- [ ] check for consistent use of  `p.TimeSteps` and `p.TimeStepsWith[Out]Grid` in new constraints
- [ ] new PR for adding "off grid" capabilities to "on grid" analyses
  - [ ] default diesel fuel cost escalation - need to document and cite (this un-does "keep "on grid" API behavior by setting the new input diesel fuel cost escalation equal to the existing input electricity cost ")
  - [ ] remove hard-coded default over-rides? primarily in reo/validator.py? maybe some in DataManager?
    - [ ] for example: critical_load_pct is set to 100 s.t. `loads_kw` is effectively used for off grid analysis; can we use the off grid flag to indicate to use `loads_kw` (instead of `critical_loads_kw`)?
  - [ ] implement tax benefits, such as where MACRS and ITC get applied in `reo/utilities.py` 
  - [ ] need all `if offgrid` logic in `reopt_model.jl`? could re-use some "on grid" code by fixing new variables like `dvLoadServed`
  - [ ] add binary cost adder for microgrids
  - [ ] add min load met percent for critical load 

- [ ] The commit message follows our guidelines


NLaws:
Note to self: if `min_turn_down_pct` is zero and there is no fuel burn y intercept for a generator then the following spinning reserve constraint:
https://github.com/NREL/REopt_Lite_API/blob/74007ed47f877792fef29030a73e9a4d81c54a37/julia_src/reopt_model.jl#L716
allows a generator to be on and providing spinning reserve at zero cost.

NLaws:
@lixiangk1 I thought that maybe one built in profile annual kWh was off because of [this test failure](https://github.com/NREL/REopt_Lite_API/runs/2975469365?check_suite_focus=true#step:9:182). So I changed it [here](https://github.com/NREL/REopt_Lite_API/pull/240/commits/51ec4b84dd7d45c13c15f67bacacc534b1091bec). However, after fixing it another built in profile [failed by 1 kWh](https://github.com/NREL/REopt_Lite_API/runs/2975843740?check_suite_focus=true#step:9:174). Is there something in this branch that is affecting the annual kWh values for the built in load profiles?

For now I will revert https://github.com/NREL/REopt_Lite_API/pull/240/commits/51ec4b84dd7d45c13c15f67bacacc534b1091bec

lixiangk1:
@NLaws Actually, yes, there was a small tweak to loads in offgrid [here](https://github.com/NREL/REopt_Lite_API/blob/6e942404540b92ba4f1bc96d8b40acb56c71b72c/reo/src/load_profile.py#L749 ). That line was initially rounding to get annual kWhs, but I changed it to a floor because it was causing infeasibilities in offgrid when meeting 100% of the load. The current julia constraint bases the load met % on annual_kwh and was failing when it got rounded up a tiny bit beyond what the load profile actually sums to.

NLaws:
@lixiangk1 I don't know why the `int` and `round` are there (@t-kwasnik added them on Jan 29 [here](https://github.com/NREL/REopt_Lite_API/commit/7e87e3f28ac508fe1300db7f5d0b2f1f694096ca)). Do you have any thoughts on how to fix this? We could either:
1. remove the `int` and `round`; or
2. change the constraint to use the sum of the load profile instead of those rounded values

julia_src/reopt_model.jl
```diff
@@ -648,6 +663,68 @@ function add_load_balance_constraints(m, p)
...
+	##Constraint (8b): Electrical Load Balancing without Grid
+	@constraint(m, ElecLoadBalanceNoGridCon[ts in p.TimeStep],
+		sum(p.ProductionFactor[t,ts] * p.LevelizationFactor[t] * m[:dvRatedProduction][t,ts] for t in p.ElectricTechs) +
+		sum( m[:dvDischargeFromStorage][b,ts] for b in p.ElecStorage )  ==
+		sum( sum(m[:dvProductionToStorage][b,t,ts] for b in p.ElecStorage) +
+			 m[:dvProductionToCurtail][t,ts]
+		for t in p.ElectricTechs) +
+        sum(m[:dvThermalProduction][t,ts] for t in p.ElectricChillers )/ p.ElectricChillerCOP +
+        sum(m[:dvThermalProduction][t,ts] for t in p.AbsorptionChillers )/ p.AbsorptionChillerElecCOP +
+		(p.ElecLoad[ts] * m[:dvLoadServed][ts])
```
NLaws, line 676: we could fix `dvLoadServed` to one in "on grid" analyses and re-use the existing load balance constraint?

lixiangk1:
@NLaws I fixed the diesel fuel esc to the elec escalation rate for the on-grid cases for now. The load test was still failing so I currently did option 2 out of what you listed to avoid having more rounding issues. Two of the CHP tests were failing for me, but those error went away locally after I merged develop, though the tests still seem to have failed after I pushed up the changes?

NLaws:
Closing so that I can be reviewer

PR #255, opened 2021-08-18
update xpress to 8.12, use reopt/xpressbase image

NLaws:
- xpress 8.12 is too large to keep in the repo (>100MB) so I created a base image with the tar.gz and install.sh files (reopt/xpressbase)
- TODO update mosel server for rancher cluster to 8.12 (this commit updates the client in julia_src/Dockerfile)
    - we will have to coordinate the upgrade to the mosel server with deploying these changes


julia_src/xpress/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:latest
+# For creating base image of Julia API. Requires an untracked tar.gz file for FICO Xpress
+# Used in both production and local development
+
+WORKDIR /opt/reopt/solver
+COPY xp8.12.3_linux_x86_64.tar.gz .
+COPY install.sh .
```
hdunham, line 7: Maybe I'm not understanding what this is doing, but how can install.sh be copied if it's been deleted?

NLaws: This Dockerfile only has to built once. Then resulting image was pushed to Docker Hub and now our team can just pull it down. I have kept this Dockerfile in the repo as a record of how the base image was built.

PR #256, opened 2021-08-19
Offgrid

adfarth:
(continuation of PR #240 ) 

## TODO
- [x] fix test suite
- [x] fill out PR template below
- [x] check for consistent use of  `p.TimeSteps` and `p.TimeStepsWith[Out]Grid` in new constraints
- [x]  Update min and max of off-grid critical_load_pct to 1.0 (in nested_inputs)
- [x] Use 'on-grid' capital cost slope instead of 'off-grid' --> remove salvage value and multiple replacements, add in tax-based incentives
- [x] Add 'replacement_year' for generator --> mirror battery behavior by defaulting to one replacement (not multiple) at the specified replacement year (only for off-grid analyses, not on-grid) 
- [x] Include diesel cost escalation for on-grid 
- [ ] new PR for adding "off grid" capabilities to "on grid" analyses
  - [ ] remove hard-coded default over-rides? primarily in reo/validator.py? maybe some in DataManager?
    - [ ] for example: critical_load_pct is set to 100 s.t. `loads_kw` is effectively used for off grid analysis; can we use the off grid flag to indicate to use `loads_kw` (instead of `critical_loads_kw`)?
  - [x] implement tax benefits, such as where MACRS and ITC get applied in `reo/utilities.py` 
  - [x] need all `if offgrid` logic in `reopt_model.jl`? could re-use some "on grid" code by fixing new variables like `dvLoadServed`
  - [ ] add binary cost adder for microgrids
  - [ ] add min load met percent for critical load 
- [ ] The commit message follows our guidelines

This PR enables modeling of off-grid microgrids by modeling a year-long grid outage. 

No off-grid specific modeling capabilities (although user could model a year-long outage without the additional capabilities listed below)

Additional capabilities include: 
- Spinning reserve requirements as a percentage of load in each timestep
- Spinning reserve requirements as a percentage of PV generation in each timestep
- Inclusion of generator fuel cost escalation % (will be applied to grid-connected analyses as well) 
- Inclusion of additional capital and annual costs (associated with microgrid development) 
- Allows for single generator replacement 
- Outputs LCOE 
- Includes nested dict of off-grid specific inputs

No. To enable off-grid capabilities, user should set off_grid_flag to True. This PR should not alter a typical, grid-connected analysis. 

### Other information:

The REopt Lite UI will be updated to reflect off-grid capabilities. User documentation will be updated as well. 

## v2 TODOs
- only create offgrid decision variables as needed
- do not run BAU for offgrid scenarios?


Bill-Becker:
@adfarth I have a project outside our group that is interested in this off grid functionality. Do you know if this is in a working state for just the electric techs?

adfarth:
> @adfarth I have a project outside our group that is interested in this off grid functionality. Do you know if this is in a working state for just the electric techs?

@Bill-Becker Yes, it is! I have been using it for a few projects. Let me know if you run into any issues or if it would be helpful for me to send along an example post. FYI I'll be working on this pull request in the next 2 weeks.

NLaws:
> (2) if there's any reason not to round sum(p.ElecLoad) down (floor) to avoid infeasibilities due to rounding?

Makes sense to me to round the value used in the constraint down to the nearest kWh. Just need to be cognizant if this affects any related outputs, such as energy costs. For example, do we need to use rounded values throughout the model so that the costs of meeting energy align with the amount of energy that has to be met?

@Bill-Becker is the best to answer:
> (1) if it definitely makes sense to use the non-cooling electric load in this case

Bill-Becker:
@NLaws @adfarth Regarding using non_cooling_electric_load, I think we are good there. This value is actually always equal to the total electric load for outages, and in the case where cooling load is not considered.

reo/src/data_manager.py
```diff
@@ -972,13 +1021,17 @@ def _get_time_steps_with_grid(self):
...
-        time_steps_with_grid = list()
-        time_steps_without_grid = list()
-        for i, pf in enumerate(self.util.prod_factor):
-            if pf > 0.5:
-                time_steps_with_grid.append(i+1)
-            else:
-                time_steps_without_grid.append(i+1)
+        if self.off_grid_flag:
+            time_steps_with_grid = []
+            time_steps_without_grid = list(range(8760)[1:])
```
NLaws, line 1026: @adfarth do you know why `time_steps_without_grid` is a range from 1 to 8759 and not 1 to 8760? 

Also, are offgrid analyses limited to an hourly time resolution?

adfarth: @NLaws It looks like currently in [validators.py](https://github.com/NREL/REopt_Lite_API/blob/ea9edd6d5215325fa0373ecfef219f00040033d5/reo/validators.py#L499) there is code to allow for multiple time steps per hour: 

`            if self.off_grid_flag:
                self.input_dict["Scenario"]["Site"]["LoadProfile"]["outage_start_time_step"] = 1
                # Previously was not modeling tax incentives in offgrid scenarios 
                # self.input_dict["Scenario"]["Site"]["Financial"]["offtaker_tax_pct"] = 0.0
                # self.input_dict["Scenario"]["Site"]["Financial"]["owner_tax_pct"] = 0.0
                if self.input_dict["Scenario"]["time_steps_per_hour"] == 4:
                    self.input_dict["Scenario"]["Site"]["LoadProfile"]["outage_end_time_step"] = 35040
                elif self.input_dict["Scenario"]["time_steps_per_hour"] == 2:
                    self.input_dict["Scenario"]["Site"]["LoadProfile"]["outage_end_time_step"] = 17520
                else:
                    self.input_dict["Scenario"]["Site"]["LoadProfile"]["outage_end_time_step"] = 8760`

But the code does break when trying to run multiple time steps per hour (likely related to the range you've referenced in data_manager.py. I will look into this and propose a solution.

adfarth: @NLaws  I think it makes sense to entirely remove the `if self.off_grid_flag and just use the else: code as below: `
```suggestion
                    time_steps_with_grid = list()
        time_steps_without_grid = list()
        for i, pf in enumerate(self.util.prod_factor):
            if pf > 0.5:
                time_steps_with_grid.append(i+1)
            else:
                time_steps_without_grid.append(i+1)
```


I tested this and it seems to be working, and can push the change.

reo/nested_inputs.py
```diff
@@ -102,6 +102,198 @@ def list_of_dict(input):
...
+    "optimality_tolerance_techs": {
+      "type": "float",
+      "min": 1.0e-5,
+      "max": 10.0,
+      "default": 0.05,
+      "description": "The threshold for the difference between the solution's objective value and the best possible value at which the solver terminates"
+    },
+    "Site": {
+      "Financial": {
+        "om_cost_escalation_pct": {
```
NLaws, line 116: @adfarth some of the `off_grid_defaults` match the existing default values (such as for `om_cost_escalation_pct`). Can we remove the definitions for these keys from `off_grid_defaults`?

reo/process_results.py
```diff
@@ -320,6 +320,77 @@ def upfront_capex_after_incentives(self):
...
+            if lcoe is not None:
+                lcc = self.nested_outputs["Scenario"]["Site"]["Financial"]["lcc_us_dollars"]
+                
+                fuel = (self.nested_outputs["Scenario"]['Site']['Generator']['total_fuel_cost_us_dollars'] or 0)
+                other_capex = (self.nested_outputs["Scenario"]["Site"]["Financial"]["additional_cap_costs_us_dollars"] or 0)
+                other_annual_costs = (self.nested_outputs["Scenario"]["Site"]["Financial"]["total_annual_cost_us_dollars"] or 0)
+                
+                analysis_period = self.inputs["Financial"]["analysis_years"]
+                discount_rate = self.inputs["Financial"]["owner_discount_pct"] # This is set to offtaker_disc_pct if third_party is false
+                # diesel_unit_cost = setup_capital_cost_offgrid(analysis_period, discount_rate,
```
NLaws, line 343: @adfarth can we remove all of the commented code in this new method? If not can you please add comments explaining what the commented code is for?

reo/nested_inputs.py
adfarth: Yes, I think we can. I left them just in case we wanted to make these defaults different from on-grid in the future, but we can always add them back in. Would you like me to remove all duplicates?

reo/process_results.py
adfarth: @NLaws  I added a comment mentioning that we may want to pull this capability back in in the future. Is that okay?

reo/nested_inputs.py
NLaws: Yes thank you

reo/process_results.py
NLaws: I try to abide by the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle, but since the code is in the commit history you could just put a note of what the capability is and the hash of the last commit that includes the code. It's fine if you want to leave it too.

reo/nested_inputs.py
adfarth: I've removed all duplicates in the commit "Remove off-grid duplicates from nested_inputs.py"

reo/process_results.py
adfarth: Ok! I removed the commented code both in process_results.py and data_manager.py

reo/process_results.py
```diff
@@ -451,6 +489,12 @@ def get_nested(self):
                     self.nested_outputs["Scenario"]["Site"][name]["sustain_hours"] = int(self.dm["LoadProfile"].get("bau_sustained_time_steps") / (len(self.dm["LoadProfile"].get("year_one_electric_load_series_kw"))/8760))
                     self.nested_outputs["Scenario"]["Site"][name]["bau_sustained_time_steps"] = self.dm["LoadProfile"].get("bau_sustained_time_steps")
                     self.nested_outputs["Scenario"]["Site"][name]['loads_kw'] = self.dm["LoadProfile"].get("year_one_electric_load_series_kw")
+                    self.nested_outputs["Scenario"]["Site"][name]["load_met_series_kw"] = self.results_dict.get("load_met")
+                    self.nested_outputs["Scenario"]["Site"][name]["load_met_pct"] = self.results_dict.get("load_met_pct")
+                    self.nested_outputs["Scenario"]["Site"][name]["sr_required_series_kw"] = self.results_dict.get("sr_required_load")
+                    #Debug variables
```
NLaws, line 495: Are these still "Debug variables"? There are in nested_outputs and saved in the database. Should we remove them or keep them as regular outputs?

NLaws (approved):
I've approved the PR and just have a couple questions. The first one is about removing the "Debug outputs" for spinning reserves. The second is related to test_offgrid: I removed all the print statements from it, but were any of those print statements being used as manual tests that should be added to the automated testing already in test_offgrid?

adfarth:
> I've approved the PR and just have a couple questions. The first one is about removing the "Debug outputs" for spinning reserves. The second is related to test_offgrid: I removed all the print statements from it, but were any of those print statements being used as manual tests that should be added to the automated testing already in test_offgrid?

I added a test to check the spinning reserve constraints. Otherwise, none of the other print statements need an automated test.

PR #257, opened 2021-08-24
Ghp

Bill-Becker:

FeatureS: This PR includes both the FY21 AOP deliverables for the AMO CHP and GTO GHP projects

1. Added /ghpghx endpoint to size the GHP(GHX) tech
a. This endpoint is called in scenario.py to size the GHP technology for input to the /reo job endpoint
b. The endpoint uses a julia_api endpoint in http.jl because the GHPGHX model is implemented in Julia
2. GHP tech
a. GHP is pre-sized and dispatched for REopt to make a binary decision on selecting GHP
3. SteamTurbine tech
a. This is exposed as a subset (prime mover) of CHP in the UI
b. The default cost and performance parameters are limited to a back-pressure style steam turbine for the AMO CHP project
4. NewBoiler tech
a. This is not exposed in the UI and referenceable default cost and performance parameters are not included
b. This technology allows a number of technologies to be modeled if the user provides appropriate inputs, including Nuclear, Biomass, Waste-to-Energy, and Geothermal direct use.
5. Supplementary firing for CHP to produce excess thermal (steam), above what would otherwise be generated from the CHP size

No.


Bill-Becker:
@NLaws Sorry in advance, this is a big PR!!

Charge code:
AMFG.10120.18.00.90

NLaws:
@Bill-Becker  is there any logical way to break up this PR? 1,634 files changed is a lot.  Are the CHP and GHP changes co-dependent?

Bill-Becker:
@NLaws The vast majority of the file changes are just the load profiles in the input_files folder because I 1) split out heating into space heating and DHW, and 2) organized that folder which makes a "change" because it's in a new folder now.

There are some changes that I was worried about conflicts if I waited too long to merge them because both project's technologies required modifications to the heating load processing and reopt_model energy balance constraints.

ghpghx/models.py
```diff
@@ -0,0 +1,300 @@
...
+from picklefield.fields import PickledObjectField
+from django.core.exceptions import ValidationError
+from django.core.validators import MinValueValidator, MaxValueValidator
+import pandas as pd
+import os
+
+log = logging.getLogger(__name__)
+
+
+class GHPGHXModel(models.Model):
```
NLaws, line 43: @Bill-Becker are you okay if I "squash" the 12 ghpgx migrations into one? I think it makes sense because the fields in the 12 files all belong to one model.

ghpghx/models.py
```diff
@@ -0,0 +1,300 @@
...
+        default=15.0, validators=[MinValueValidator(0.0), MaxValueValidator(100.0)],
+        help_text="Pumping power required for a given volumetric flow rate of the fluid in the GHX [Watt/GPM]")
+    ghx_pump_min_speed_fraction = models.FloatField(null=True, blank=True, 
+        default=0.1, validators=[MinValueValidator(0.0), MaxValueValidator(1.0)],
+        help_text="The minimum turndown fraction of the pump. 1.0 is a constant speed pump.")
+    ghx_pump_power_exponent = models.FloatField(null=True, blank=True, 
+        default=2.2, validators=[MinValueValidator(0.1), MaxValueValidator(10.0)],
+        help_text="The pump power curve exponent")
+    max_eft_allowable_f = models.FloatField(null=True, blank=True, 
+        default=104.0, validators=[MinValueValidator(0.0), MaxValueValidator(150.0)],
```
NLaws, line 122: @Bill-Becker Any field that has a default value should not have `null=True` (which implies that a null value can be saved to the database). Please remove `null=True` from all values with defaults.

ghpghx/models.py
```diff
@@ -0,0 +1,300 @@
...
+                    "You may have hit the results endpoint too quickly after POST'ing scenario, "
+                    "you may have a typo in your ghp_uuid, or the scenario was deleted.").format(ghp_uuid)
+                resp['status'] = 'error'
+                return resp
+            else:
+                raise Exception
+        #ghpghx_data = remove_ids(model_to_dict(ghpghx_model))
+        ghpghx_data = model_to_dict(ghpghx_model)
+
+        # Listed here to separate outputs from the model, so need to keep this update with added outputs to model
```
NLaws, line 275: We would not have to keep the `output_fields` updated if we split the GHPGHXModel into its inputs and outputs (like in v2 job/models). Should we split the model into two?

ghpghx/process_ghpghx_results.py
```diff
@@ -0,0 +1,135 @@
...
+import copy
+import numpy as np
+import logging
+from ghpghx.models import GHPGHXModel, ModelManager
+log = logging.getLogger(__name__)
+
+
+#TODO Update this to use celery for task queuing /ghpghx jobs (or wait until V2 to parallelize in Julia)
+
+def process_ghpghx_results(data, meta, saveToDB=True):
```
NLaws, line 41: looks this method is never used? is it something that you are saving for the TODO above it?

reo/scenario.py
```diff
@@ -317,29 +337,108 @@ def setup_pv(pv_dict, latitude, longitude, time_steps_per_hour):
...
+                        ghpghx_post["heating_fuel_load_mmbtu_per_hr"] = lpbf_space.load_list
+                    else:
+                        ghpghx_post["heating_fuel_load_mmbtu_per_hr"] = [lpbf_space.load_list[i] + lpbf_dhw.load_list[i] for i in range(len(lpbf_space.load_list))]
+                if dfm.boiler is not None:
+                    ghpghx_post["existing_boiler_efficiency"] = dfm.boiler.boiler_efficiency #boiler.boiler_efficiency
+                else:
+                    ghpghx_post["existing_boiler_efficiency"] = 0.8
+                if ghpghx_post.get("cooling_thermal_load_ton") in [None, []]:
+                    ghpghx_post["cooling_thermal_load_ton"] = [kwt / TONHOUR_TO_KWHT for kwt in dfm.cooling_load.load_list] #lpct.load_list
+                client = TestApiClient()
```
NLaws, line 381: @Bill-Becker Why not just use the ghpghx functions directly? And/or call the Julia API? I'm wary of using a testing tool in production.

reo/scenario.py
```diff
@@ -317,29 +337,108 @@ def setup_pv(pv_dict, latitude, longitude, time_steps_per_hour):
...
+                #json.dump(ghpghx_response_list, open("ghpghx_response.json", "w"))
+                ghp_option_list.append(ghp.GHPGHX(dfm=dfm,
+                                                    response=ghpghx_response_list[i],
+                                                    **inputs_dict["Site"]["GHP"]))
+            # Update GHPModel with created ghpghx_response
+            tmp = dict()
+            tmp['ghpghx_response'] = ghpghx_response_list
+            ModelManager.updateModel('GHPModel', tmp, run_uuid)
+            # Sleep to avoid calling julia_api for /job (reopt) too quickly after /ghpghx (only required locally and/or for debugging?)
+            time.sleep(10)
```
NLaws, line 404: Why is this long sleep needed?

reo/scenario.py
```diff
@@ -317,29 +337,108 @@ def setup_pv(pv_dict, latitude, longitude, time_steps_per_hour):
...
-                    
+
         dfm.add_soc_incentive = inputs_dict['add_soc_incentive']
 
+        if inputs_dict["Site"]["NewBoiler"]["max_mmbtu_per_hr"] > 0:
+            newboiler = NewBoiler(dfm=dfm, **inputs_dict['Site']['NewBoiler'])
+            # Any parameters processed and updated?
+            tmp = dict()
+            # Assign tmp["param"] = newboiler.xyx
+            ModelManager.updateModel('NewBoilerModel', tmp, run_uuid)
```
NLaws, line 425: Why update the model with an empty `dict`? (same for SteamTurbine).

reo/process_results.py
```diff
@@ -248,6 +250,11 @@ def upfront_capex(self):
                 elif len(cost_list) == 1:
                     upfront_capex += (cost_list[0] or 0) * (chp_size or 0)
 
+            # GHP
+            if len(self.dm["ghp_cost"]) >= 1:
+                # TODO currently this only reports the first option cost
```
NLaws, line 255: Looks like a TODO that needs addressed before production?

julia_src/ghpghx_results.jl
```diff
@@ -0,0 +1,104 @@
+
+Base.@kwdef mutable struct ResultsStruct
```
NLaws, line 2: Do the results need to be mutable? It is safer to keep any data struct immutable.

julia_src/GHPGHX.jl
```diff
@@ -0,0 +1,463 @@
...
+    # Load parameters and process from input Dict, to create instance of type InputsStruct
+    p = InputsProcess(d)
+
+    # Run size_borefield for each attempt to SOLVE
+    results = size_borefield(p)
+
+    return results, p
+end
+
+function size_borefield(p)
```
NLaws, line 20: Can you please add a docstring to this function and the others giving a high level summary of what each function does?

julia_src/ghpghx_inputs.jl
```diff
@@ -0,0 +1,251 @@
...
+    solver_eft_tolerance_f::Float64  # Tolerance for the EFT error to accept a GHX sizing solution
+    solver_eft_tolerance::Float64  # Convert to degC
+    ghx_model::String  # "TESS" or "DST"
+    dst_ghx_timesteps_per_hour::Int64
+    tess_ghx_minimum_timesteps_per_hour::Int64
+    max_sizing_iterations::Int64
+    init_sizing_factor_ft_per_peak_ton::Float64
+    
+    ##### These are the variable names used in the GHPGHX, kept from TESS ######
+    # TODO eventually just use the API names above in GHPGHX to remove redundancy (BUT WOULD HAVE TO DEAL WITH UNITS CONVERSION STILL)
```
NLaws, line 48: It would be really nice to have naming consistency.

julia_src/Dockerfile
```diff
@@ -24,6 +24,10 @@ ARG XPRESS_LICENSE_HOST=licenseserver
 RUN set -x && printf "f\nc\n\nn\n\n${XPRESS_LICENSE_HOST}\ny\n" | ./install.sh >> license_info.txt;
 RUN rm xp8.12.3_linux_x86_64.tar.gz
 
+# Install gfortran for /ghpghx
+RUN apt-get update
```
NLaws, line 28: This line and the next should be one `RUN` command (see https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run).

ghpghx/resources.py
```diff
@@ -0,0 +1,207 @@
...
+        data["inputs"] = ModelManager.make_response(ghp_uuid=ghp_uuid)["inputs"]
+        # Remove extra inputs used above but not expected in ghpghx_inputs.jl
+        data["inputs"].pop("latitude", None)
+        data["inputs"].pop("longitude", None)
+        data["inputs"].pop("heating_fuel_load_mmbtu_per_hr", None)
+        data["inputs"].pop("existing_boiler_efficiency", None)
+        data["status"] = 'Solving for GHX Size...'
+
+        try:
+            # TODO Make this a celery task so we don't have to wait before getting ghp_uuid back in response
```
NLaws, line 173: @Bill-Becker do you still plan on making the ghpghx call a celery task? (If not it looks like the run_ghpghx_sizing.py and process_ghpghx_results.py files are obsolete.)

ghpghx/urls.py
```diff
@@ -0,0 +1,41 @@
...
+# LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE
+# OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+# OF THE POSSIBILITY OF SUCH DAMAGE.
+# *********************************************************************************
+from django.conf.urls import url, re_path
+from django.conf.urls.static import static
+from django.conf import settings
+from . import views
+
+urlpatterns = [
```
NLaws, line 35: Just curious: do you expect eternal use of these endpoints and the main GHP POST endpoint? If not we could simplify matters be not serving these URLs.

ghpghx/validators.py
```diff
@@ -0,0 +1,39 @@
...
+# OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+# OF THE POSSIBILITY OF SUCH DAMAGE.
+# *********************************************************************************
+import logging
+from ghpghx.models import GHPGHXModel
+from django.core.exceptions import ValidationError
+log = logging.getLogger(__name__)
+
+"""
+Define reusable/importable django model validators to be used in
```
NLaws, line 36: Unneeded file? Or a TODO that needs addressed?

julia_src/reopt_model.jl
```diff
@@ -195,15 +223,15 @@ function add_bigM_adjustments(m, p)
 	for n in p.DemandMonthsBin
 		for mth in p.Month
 			if n > 1
-				m[:NewMaxDemandMonthsInTier][mth,n] = minimum([p.MaxDemandMonthsInTier[n],
-					added_power + 2*maximum([p.ElecLoad[ts] + p.CoolingLoad[ts] / p.ElectricChillerCOP
-					for ts in p.TimeStepRatchetsMonth[mth]])  -
-					sum(m[:NewMaxDemandMonthsInTier][mth,np] for np in 1:(n-1)) ]
-				)
+				m[:NewMaxDemandMonthsInTier][mth,n] = minimum([p.MaxDemandMonthsInTier[n], DemandBigM])
+				# 	added_power + 2*maximum([p.ElecLoad[ts] + p.CoolingLoad[ts] / p.ElectricChillerCOP
```
NLaws, line 227: there are a lot of commented out lines in `add_bigM_adjustments`: why is that?

julia_src/utils.jl
```diff
@@ -221,7 +221,7 @@ Base.@kwdef struct Parameter
 	BoilerTechs::Array{String,1}
 	HeatingLoad::Array{Float64,1}
 	CoolingLoad::Array{Float64,1}
-	BoilerEfficiency::Float64
+	BoilerEfficiency::AxisArray
```
NLaws, line 224: I have been finding AxisArray's somewhat problematic (such as when assigning a vector of values to them) and so I have been switching them to Dict's. The `BoilerEfficiency` inputs seems like a natural for a Dict: string keys for numeric values. Not necessary to change now (we can do it with v2) so just an FYI.

reo/migrations/0107_merge_20210420_1418.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 3.1.8 on 2021-04-20 14:18
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
NLaws, line 6: We could probably squash all of these migrations into one as well, or make one migration for each new model.

reo/nested_inputs.py
```diff
@@ -289,7 +296,7 @@ def list_of_dict(input):
         "annual_kwh": {
           "type": "float",
           "min": 1.0,
-          "max": max_big_number,
+          "max": 1.0E10,
```
NLaws, line 299: Why did the `annual_kwh` max input need to be increased? (Maybe we should consider scaling our problems to megawatts?).

reo/nested_inputs.py
```diff
@@ -1658,7 +1708,11 @@ def list_of_dict(input):
           "min": 0.0,
           "max": 1.0,
           "description": "Knockdown factor on absorption chiller COP based on the CHP prime_mover not being able to produce as high of temp/pressure hot water/steam"
-        }
+        },
+        "can_supply_st": {
```
NLaws, line 1712: Why not spell out "steam" in this input name? I think that we should to make it clearer.

reo/nested_inputs.py
```diff
@@ -1833,6 +1907,307 @@ def list_of_dict(input):
...
+        "macrs_bonus_pct": {
+          "type": "float",
+          "min": 0.0,
+          "max": 1.0,
+          "default": 0.0,
+          "description": "Percent of upfront project costs to depreciate under MACRS"
+        }        
+      },
+      "GHP": {
+        "force_ghp": {
```
NLaws, line 2063: "force" seems like the wrong term: maybe "require_ghp_purchase" or "require_ghp_install" ?

reo/nested_inputs.py
```diff
@@ -1833,6 +1907,307 @@ def list_of_dict(input):
...
+          "type": "float",
+          "min": 0.0,
+          "max": max_big_number,
+          "description": "Building square footage for GHP/HVAC cost calculations"
+        },
+        "ghpghx_inputs": {
+          "type": "list_of_dict",
+          "description": "GHP-GHX inputs/POST to /ghpghx endpoint"
+        },
+        "ghpghx_response": {
```
NLaws, line 2113: is "ghpghx_response" necessary? Maybe just provide the id to get the response so that we aren't saving the same values in two places.

reo/src/data_manager.py
```diff
@@ -406,16 +487,16 @@ def _get_REopt_cost_curve(self, techs):
                     if u == 0 or u_cap == 0:
                         switch_rebate = True
 
-                    # start at second point, first is always zero
+                    # TODO not true: start at second point, first is always zero
```
NLaws, line 490: "not true" because of affine cost curves? Do we need to address this now?

reo/src/data_manager.py
```diff
@@ -1216,6 +1321,42 @@ def finalize(self):
...
+                ghp_size_tons = option.heatpump_capacity_tons
+                if ghp_size_tons <= ghp_cap_cost_x[1]:
+                    n = 0
+                elif ghp_size_tons > ghp_cap_cost_x[-1]:
+                    n = ghp_n_segments - 1
+                else:
+                    for n in range(ghp_n_segments):
+                        if (ghp_size_tons > ghp_cap_cost_x[n]) and (ghp_size_tons <= ghp_cap_cost_x[n+1]):
+                            break
+                # TODO make sure the below is consistent with the y-int + slope * size definitions
```
NLaws, line 1345: Is this TODO is related to the cost curve calculations above? Or the constraints in the JuMP model? Maybe needs to be addressed.

reo/src/ghp.py
```diff
@@ -0,0 +1,119 @@
...
+from reo.src.incentives import Incentives, IncentivesNoProdBased
+from reo.src.data_manager import big_number
+
+
+class GHPGHXInputs:
+    """
+    Map the GHPGHX inputs to the objects in this class
+    """
+
+    ground_k_by_climate_zone = {
```
NLaws, line 14: It looks like the only use of the GHPGHXInputs class is the `ground_k_by_climate_zone` attribute. You have a TODO below about adding all of the input fields. Maybe we can settle on one input model? We could start by moving `ground_k_by_climate_zone` to the ghpghx/models.py perhaps.

reo/src/ghp.py
```diff
@@ -0,0 +1,119 @@
+import numpy as np
+import pandas as pd
```
NLaws, line 2: pandas is not used in this file. many of the new files contain unused imports. please clean them up (probably a tool that can do it for you).

reo/src/ghp.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def __init__(self, inputs_dict, **kwargs):
+        #TODO Add ALL the input fields for the GHPGHX model
+        #   (remove all intermediate calculated fields based on inputs from here, so it's cleaner as the GHP POST)
+        self.heating_thermal_mmbtu_per_hr = inputs_dict["heating_thermal_load_mmbtu_per_hr"]
+        self.heating_thermal_kw = list(np.array(self.heating_thermal_mmbtu_per_hr) * MMBTU_TO_KWH)
+        self.cooling_thermal_ton = inputs_dict["cooling_thermal_load_ton"]
+        self.cooling_thermal_kw = list(np.array(self.cooling_thermal_ton) * TONHOUR_TO_KWHT)
+
+class GHPGHXOutputs:
```
NLaws, line 40: Since the class only has an init method it is only a data filter: it would be much more efficient to replace it with a function. However, is it even necessary (as a function)? Why not just use `outputs_dict` where ever GHPGHXOutputs is used?

reo/src/load_profile.py
```diff
@@ -40,6 +40,18 @@
...
+library_path_base = os.path.join('input_files', 'LoadProfiles')
+load_type_file_map = {"Electric": "Load8760_norm_",
+                        "SpaceHeating": "SpaceHeating8760_norm_",
+                        "DHW": "DHW8760_norm_",
+                        "Cooling": "Cooling8760_norm_"}
+
+space_heating_annual_loads = json.load(open(os.path.join(library_path_base, "space_heating_annual_mmbtu.json"), "rb"))
+dhw_annual_loads = json.load(open(os.path.join(library_path_base, "dhw_annual_mmbtu.json"), "rb"))
+total_heating_annual_loads = {city: {building: space_heating_annual_loads[city][building] + dhw_annual_loads[city][building] for building in space_heating_annual_loads[city].keys()} for city in space_heating_annual_loads.keys()}
+space_heating_fraction_flat_load = {city: space_heating_annual_loads[city]["flatload"] / total_heating_annual_loads[city]["flatload"] for city in space_heating_annual_loads.keys()}
```
NLaws, line 54: Since all of this data on lines 51-54 are loaded in when anything from this file is imported (i.e. when the API is started) then all of this data consumes RAM for all time. It looks like at least some of the data is only used conditionally so it would make more sense to only load the data conditionally - unless it is a speed bottleneck.

reo/src/load_profile.py
```diff
@@ -777,3 +817,30 @@ def _account_for_existing_pv(self, pvs, years):
             native_load = [i + j for i, j in zip(self.unmodified_load_list, existing_pv_kw_list)]
             return native_load, existing_pv_kw_list
         return copy.copy(self.unmodified_load_list), existing_pv_kw_list
+
+def get_climate_zone(latitude, longitude):
```
NLaws, line 821: It looks we could use this function in `BuiltInProfile.city` to cut down on redundant code: maybe we should move `get_climate_zone` to utilities.py? I can do this if you agree.

reo/src/load_profile_boiler_fuel.py
```diff
@@ -1,4 +1,5 @@
-from reo.src.load_profile import BuiltInProfile
+from reo.src.load_profile import library_path_base, BuiltInProfile, space_heating_annual_loads, dhw_annual_loads, \
```
NLaws, line 1: There are a lot of unused imports in this file.

reo/urls.py
```diff
@@ -44,5 +44,6 @@
     re_path(r'^loadprofile_chillerthermal_chiller_cop/?$', views.loadprofile_chillerthermal_chiller_cop),
     re_path(r'^absorption_chiller_defaults/?$', views.absorption_chiller_defaults),
     re_path(r'^schedule_stats/?$', views.schedule_stats),
+    re_path(r'^ground_conductivity/?$', views.ground_conductivity),
```
NLaws, line 47: The ground_conductivity endpoint seems like it would be better housed in the ghp app, but do we need to provide it as an endpoint at all? Seems like the REopt API should not be the (public) source for ground conductivity.

NLaws (changes requested):
@Bill-Becker I did a first pass through and added many inline comments. At a high level my proposed changes are to clean up and clarify the GHP inputs/outputs from the main `job` endpoint and the `ghpghx` endpoints. Also, I wonder how many of the new endpoints that we really need. It seems like the ghpghx app is really just internal functions that are not necessary to expose, unless the REopt API is meant to provide the GHP functions as a stand-alone capability? Also, if the ghpghx sizing problem is time consuming then we should make them into celery tasks.

Bill-Becker:
@NLaws Thanks for the review. I'll go through an address the comments soon. I agree we can improve the inputs/outputs for ghpghx with separated models like V2 and with respect to the job endpoint. We do plan to run /ghpghx independent of /job. It's a nice way to get one or more `ghpghx_response` entries to use as inputs for /job runs; i.e. you can use the same `ghpghx_response` input for many /job runs which avoids having to run the ghpghx model for each /job run. With additional funding that might come from GTO, the vision was to create a "GHP builder" (like the Rate Builder) in the webtool so users could run /ghpghx separately and have their own stored database of `ghpghx_response`s to choose from in a /job evaluation. Also, the SAM team expressed interest in GHP (and we expect others possibly), and /ghpghx would allow external apps to leverage the detailed ghpghx sizing and performance tool that we developed to be integrated into their apps. We also hope to get follow-on funding to do some more interesting GHP system configurations which will require some different inputs/outputs interactions with the /job endpoint.

ghpghx/models.py
Bill-Becker: Done

ghpghx/models.py
Bill-Becker: This is a good idea and should be done. But, I'm going to punt on this until I convert GHP to V2 when I'm in the process of converting all the other thermal technologies over.

Bill-Becker: There may be a theme in my remaining responses too, that I'll try to address everything that is required now, but I may punt on "nice to haves" because of time crunch now. We are pushing off the webtool public release until Q1 FY22, and I'll have more time to address the nice-to-haves then as well.

ghpghx/process_ghpghx_results.py
Bill-Becker: Yes, as you suggest I removed these files that were not used because we are not creating celery tasks for them.

ghpghx/resources.py
Bill-Becker: These files have been removed.

reo/scenario.py
Bill-Becker: This way leverages all the same code in `/ghpghx/resources.py` for inputs validation, POSTing to julia_api (http.jl /ghpghx endpoint), error handling, and creating and saving the ghpghx db models. This has been in the production environment for several months now with 1000+ runs, and I haven't seen any issues with it. Let me know if you want to discuss alternative solutions.

reo/scenario.py
Bill-Becker: I think it's unnecessarily long, but while testing I was getting some julia_api connection errors when I didn't have any sleep. I thought it had to do with POSTing to the julia_api /job endpoint too soon after POSTing to the julia_api /ghpghx endpoint. I just commented out the sleep and I didn't get any connection errors, so maybe we don't need it. Just to be safe and if you're OK with it, I'm going to reduce it to 1 second.

Bill-Becker: Actually I just realized I left a note in there that it was giving me a connection error when sequentially running multiple julia_api /ghpghx endpoints back-to-back in the for-loop for the number of ghpghx_inputs that were passed. Either way, I'm not getting that connection error now, so not sure. Maybe my computer was bogged down then.

reo/scenario.py
Bill-Becker: I removed this updateModel from both

reo/process_results.py
Bill-Becker: Did the TODO

julia_src/ghpghx_results.jl
Bill-Becker: Yes! This struct gets updated at every iteration (sub-hourly typically) of the ghpghx simulation period by the GHX Fortran subroutine. For each sizing iteration in the sizing solver, it gets reset with the zero values assigned using the kwargs functionality of this struct.

julia_src/GHPGHX.jl
Bill-Becker: Added for all of the /ghpghx related .jl functions and structs.

julia_src/ghpghx_inputs.jl
```diff
@@ -0,0 +1,251 @@
+# Using same struct type from REopt utils.jl
+# Warning: I've seen notes of Base.@kwdef not being officially supported (not documented)
```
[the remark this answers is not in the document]
Bill-Becker: Would that also require changing the way the [InputsProcess](https://github.com/NREL/REopt_Lite_API/blob/fdd35af95007f2551102fa6b030be99e6b975c7f/julia_src/ghpghx_inputs.jl#L113) [constructor?] works? I admit I'm still a little fuzzy on structs in general and especially best-practices for updating/modifying them using functions.

julia_src/ghpghx_inputs.jl
Bill-Becker: I agree! But, I think I'm going to categorize this as a "nice-to-have" and keep this as a TODO for now. I will tackle this next FY with the follow-on GTO funds when we do a little more development on this GHPGHX.jl module and I can align GHP with V2.

julia_src/Dockerfile
Bill-Becker: Done

ghpghx/resources.py
Bill-Becker: No, I removed this comment.

ghpghx/urls.py
Bill-Becker: We do want the ghpghx.resources GHPGHXJob (POST-style) and the /results available externally. The /help url is a TODO, and I removed the other one (/post_exmpl) that was not being used.

ghpghx/validators.py
Bill-Becker: I'd like to leave this file in here because it will be used eventually to add some robustness to ghpghx inputs validation by doing those TODO notes and importing those custom validators into models.py and resources.py (GHPGHXJob).

julia_src/reopt_model.jl
Bill-Becker: The commented-out bigM heuristics don't work for GHP because we are electrifying the heating load, so the electric load demand/usage bigM can't be based on the BAU electric load alone. I've effectively replaced the heuristic NewMaxDemand/Usage with a big number (if there are not tiers, otherwise use the tiered parameters).

Bill-Becker: I just uncommented out the heuristic bigM calcs and added GHP-specific adjustments to the calculations, so this preserves the previous logic. I've added the heating load to the electric-based bigM calcs if GHP is considered (implicitly assumes a worst-case COP of 1 for GHP), and I removed the electric chiller COP from the CoolingLoad term in case the GHP COP is lower than the BAU electric chiller COP.

julia_src/utils.jl
Bill-Becker: Ok, noted, and yes we can convert with V2 (there are a lot of those for CHP)

reo/migrations/0107_merge_20210420_1418.py
Bill-Becker: Ok, I can do that. Do you know if that will require undeploying and redeploying the branch on staging to avoid db errors?

reo/nested_inputs.py
Bill-Becker: I think we had a large DoD site that exceeded the ~11.4 MW average load, so I increased the max number up by 2 orders of magnitude to be safe.

reo/nested_inputs.py
Bill-Becker: st actually = "steam turbine", so it's kind of an acronym. However, your point is still valid and I think it's not clear enough - I was just being concise but at the expense of clarity. I'll rename to `can_supply_steam_turbine`.

reo/nested_inputs.py
Bill-Becker: I agree it is saving it in two places, but I'd like to keep it this way at least for now so that the reo/job runs with GHP can be re-created/copied without dependence on the state of a particular database.

reo/src/data_manager.py
Bill-Becker: I don't remember the fix I did for GHP with "first is always zero", but I worked around the existing framework. So I'm deleting the TODO

reo/src/data_manager.py
Bill-Becker: I did "make sure" and just forgot to remove this, so I'm removing the TODO

NLaws:
@Bill-Becker Please let me know when you are ready for me to go through this PR again. Also, could you please add a README.md to input_files/LoadProfiles describing the break down and units of the loads? For example: is DHW the load for gas or electric  heaters or both and what are the appropriate units to scale them with? (It looks like the units are MMBtu because there is a file `dhw_annual_mmbtu.json` but it would be nice to make it clear). And, on the file `dhw_annual_mmbtu.json` and others: please include in the README where the values come from. Do we have scripts some where for creating these? What is annual_heating_stats.csv? etc.

julia_src/reopt_model.jl
```diff
@@ -509,21 +599,38 @@ function add_thermal_load_constraints(m, p)
...
-				sum(m[:dvProductionToWaste][t,ts] for t in p.CHPTechs) + sum(m[:dvProductionToStorage][b,t,ts] for b in p.HotTES, t in p.HeatingTechs)  +
-				sum(m[:dvThermalProduction][t,ts] for t in p.AbsorptionChillers) / p.AbsorptionChillerCOP
-		)
+        if !isempty(p.SteamTurbineTechs)
+            @constraint(m, HotThermalLoadCon[ts in p.TimeStep],
+                    sum(m[:dvThermalProduction][t,ts] - m[:dvThermalToSteamTurbine][t,ts] for t in p.CHPTechs) +
+                    sum(m[:dvThermalProduction][t,ts] for t in p.SteamTurbineTechs) +
+                    sum(p.ProductionFactor[t,ts] * (m[:dvThermalProduction][t,ts] - m[:dvThermalToSteamTurbine][t,ts]) for t in p.BoilerTechs) +
+                    sum(m[:dvDischargeFromStorage][b,ts] for b in p.HotTES) ==
+                    p.HeatingLoad[ts] * p.BoilerEfficiency["BOILER"] -
```
NLaws, line 616: @Bill-Becker why is the `HeatingLoad` _multiplied_ by the (existing) boiler efficiency in the load balance? If it were just the boiler I would expect the load to be _divided_ by the boiler efficiency; but there are other techs that can meet the load so why apply only the boiler efficiency?

Bill-Becker: @NLaws HeatingLoad is _fuel_, and this is a _thermal_ energy balance, so the load has to be multiplied by efficiency.

reo/src/data_manager.py
```diff
@@ -1216,6 +1321,40 @@ def finalize(self):
...
+                    for n in range(ghp_n_segments):
+                        if (ghp_size_tons > ghp_cap_cost_x[n]) and (ghp_size_tons <= ghp_cap_cost_x[n+1]):
+                            break
+                ghp_installed_cost.append(ghp_cap_cost_yint[n] + ghp_size_tons * ghp_cap_cost_slope[n])
+                ghp_om_cost_year_one.append(option.om_cost_year_one)
+                if self.boiler is not None:
+                    heating_thermal_load = np.array(heating_load) * self.boiler_efficiency
+                else:
+                    heating_thermal_load = np.array(heating_load) * 0.8
+                ghp_heating_thermal_load_served_kw.append(list(np.minimum(option.heating_thermal_load_served_kw, heating_thermal_load)))
```
NLaws, line 1350: @Bill-Becker I thought that GHP could only meet space heating (or was it only DHW?): the code appears to be setting the `ghp_heating_thermal_load_served_kw` to the minimum of an input and the _sum_ of the space heating and DHW.

Bill-Becker: @NLaws Yes, whether GHP serves space heating (default) or the whole heating load (non-default) is set in scenario.py based on `GHP.can_serve_dhw`. This np.minimum is for rounding differences from the Julia GHPGHX.jl module that could result in GHP serve **slightly more** than the **total** heating load (talking about very small fractions which would make the problem infeasible from the thermal energy balance constraint).

ghpghx/models.py
Bill-Becker: Done

ghpghx/models.py
Bill-Becker: I ended up splitting these up per your suggestion.

reo/migrations/0107_merge_20210420_1418.py
Bill-Becker: I squashed the migrations when I split the GHPGHX models into Inputs and Outputs

reo/nested_inputs.py
Bill-Becker: I changed this ghpghx_response to be ghpghx_response_uuids so we are just storing the ghp_uuids in the reo GHP inputs dict.

reo/src/ghp.py
Bill-Becker: I moved this data to the /ghpghx endpoint. The /ground_conductivity endpoint is now served in a /ghpghx app view.

reo/src/ghp.py
Bill-Becker: Removed unused imports here and some elsewhere.

reo/src/load_profile.py
Bill-Becker: I made these conditionally loading and reduced the amount of loading and calcs by creating a couple of .json files with the data.

reo/src/load_profile.py
Bill-Becker: I did this.

reo/urls.py
Bill-Becker: I moved this into the /ghpghx app views and exposed via /ghpghx urls.

Bill-Becker:
@NLaws I believe I've addressed all of your comments. I ended up doing more than I thought I would have time for on some of your items, which is all for the better. Also, you may have noticed that Alex Z. made a PR into this branch after you reviewed for adding a cost to the CHP supplementary firing; this turned out to be more additions to reopt_model.jl than I thought, but he did a good job being very thorough and adding tests so I'm confident in those additions. 

Could you please re-review by this week? I'm supposed to have the AMO CHP related things published by end of FY (this week). Also, I accidentally chose master as the branch to merge into, but it should be develop - I didn't want to change it to develop because it said I might lose comments, so I'll change it to develop after you "approve").

Bill-Becker:
@NLaws as mentioned in this [this PR 265](https://github.com/NREL/REopt_Lite_API/pull/265), all the comments I responded to today are actually contained in that separate branch. If you'd prefer me to just merge that into `ghp` and keep your review here, let me know.

reo/tests/test_chp.py
```diff
@@ -215,3 +215,78 @@ def test_chp_resilience_unavailability(self):
...
+        c = nested_to_flat_chp(d['outputs'])
+        d_expected = {}
+        d_expected['chp_kw'] = 800
+        d_expected['chp_supplementary_firing_kw'] = 0
+        d_expected['chp_year_one_electric_energy_produced_kwh'] = 800*8760
+        d_expected['chp_year_one_thermal_energy_produced_mmbtu'] = 800*(0.4418/0.3573)*(kw_to_mmbtu)*8760
+        try:
+            check_common_outputs(self, c, d_expected)
+        except:
+            print("Run {} expected outputs may have changed. Check the Outputs folder.".format(run_uuid))
```
NLaws, line 266: We have not had an outputs folder for a long time. I know that this is just a copy and paste but we should get rid of these messages throughout the tests. In fact, I noticed that you added to the `nested_to_flat` function for some tests (or maybe Alex did), but I also think that we should deprecate `check_common_outputs`, which will mean that we no longer have to maintain `nested_to_flat`.

NLaws (approved):
Please read through my comments. Most are just comments or "nice to haves" except defining `p_cap` (I still don't know what all those terms are in the cost curve function :))

Bill-Becker:
@NLaws Thanks for the review and approval! I made some of the small suggestions (including p_cap definition, but I generally agree this is confusing logic and naming). I'm going to change this PR to merge into `develop` instead of `master`, and I'll immediately open a PR for merging `develop` into `master`.

PR #262, opened 2021-09-18
Add supplementary firing costs

zolanaj:

Feature: this PR adds a capital cost to the supplementary duct firing system that allows CHP systems to produce more heat than that which is captured via the prime mover when generating electricity.  

- Supplementary firing is included and incurs a cost according to the thermal efficiency of the system, but no capital cost for this add-on exists.
- Supplementary firing thermal production is not included in thermal output summary.

- Capital cost per kW of the supplementary firing system has been added as a new input, and the size of the supplementary firing system in the solution has been added as an output.
- A new binary variable and continuous variable have been added to the optimization model; the former is 1 if supplementary firing is included with the CHP system and 0 otherwise, while the second represents the size of the supplementary firing system.
- A collection of constraints have been added so that the size of the supplementary firing system is equal to the CHP system size if any supplementary firing is used, and zero otherwise.
- Supplementary firing thermal production is combined with the CHP prime mover heat capture in tracking CHP thermal production as a combined statistic (It's not currently separated in the outputs).

No

### Other information:
- Capital cost calculation is included in the initial capex costing when processing outputs.
- Thermal production from supplementary firing isn't tracked separately anywhere in the Julia outputs right now.


Bill-Becker:
@zolanaj Thanks for doing a very thorough job adding this feature to the API! It probably took you longer than expected. I just made one name change: `dvSupplementaryThermalSize` changed to `dvSupplementaryFiringCHPSize` to clarify that it's not a thermal size of the supplementary firing but instead the CHP size used for supplementary firing.

Bill-Becker:
@zolanaj Please review the last two commits for the addition of one constraint and the reduction of 2 others into one indicator-style constraint (we are trying to use those where applicable).

Bill-Becker (commented):
Approved!

zolanaj:
@Bill-Becker thanks for these!  Your updates look good to me, and I'll implement indicator constraints in this format moving forward.

PR #270, opened 2021-10-25
add Xpress.postsolve to http.jl (release memory?)

NLaws:
will also be covered in Xpress.jl eventually
https://github.com/jump-dev/Xpress.jl/pull/157


Bill-Becker:
@NLaws I'm reverting this commit so that we can push `offgrid` features into master, and I will re-create this PR for this commit.

PR #278, opened 2021-12-14
Ghp updates

Bill-Becker:
- Add input and defaults to account for HVAC efficiency gains from GHP retrofit which may reduce the heating and/or cooling thermal loads. Default data is from building energy modeling.
- Option to include ghpghx_response dictionary instead of relying on ghpghx database with ghp_uuid
- Add ghpghx output of ground temperature
- Add test for addressable load, space heating fraction, space heating efficiency thermal factor, cooling efficiency thermal factor


Bill-Becker:
@NLaws Changing the base branch to master to expedite the merge.

reo/validators.py
```diff
@@ -1881,6 +1882,43 @@ def check_special_cases(self, object_name_path, template_values=None, real_value
...
+                    if heating_factor in [[], None] or cooling_factor in [[], None]:
+                        latitude = self.input_dict['Scenario']['Site']['latitude']
+                        longitude = self.input_dict['Scenario']['Site']['longitude']
+                        climate_zone, nearest_city, geometric_flag = get_climate_zone_and_nearest_city(latitude, longitude, BuiltInProfile.default_cities)
+                        heating_factor_data = pd.read_csv(os.path.join('input_files', 'LoadProfiles', 'ghp_heating_efficiency_thermal_factors.csv'), index_col="Building Type")
+                        cooling_factor_data = pd.read_csv(os.path.join('input_files', 'LoadProfiles', 'ghp_cooling_efficiency_thermal_factors.csv'), index_col="Building Type")
+                        building_type_heating = self.input_dict['Scenario']['Site']['LoadProfileBoilerFuel'].get('doe_reference_name') or []
+                        building_type_cooling = self.input_dict['Scenario']['Site']['LoadProfileChillerThermal'].get('doe_reference_name') or []
+
+                    # Default thermal factors are assigned for certain building types and not for campuses (multiple buildings)
```
NLaws, line 1903: would it be too much to define the heating factor for a group of buildings as a weighted sum based on the annual_kwh? I would think that GHP systems for groups of buildings might be a common use case.

NLaws:
> @NLaws Changing the base branch to master to expedite the merge.

This branch will still need to be merged into develop and it would be good to test its deployment on testranch first.

reo/validators.py
Bill-Becker: @NLaws That is a good idea, and I plan to do that, but I'd like to leave that as a future addition due to time constraints here with Andrew then having to interface with the endpoint for getting these defaults based on multiple reference buildings. I'd also want to run that buy the buildings people. For now, the users can enter their own custom thermal factors for campuses.

NLaws (approved):
Please rebase to develop branch, merge and deploy to testranch. You can merge to master w/o approval after that.

PR #287, opened 2021-12-30
Divbyzero fix

adfarth:

Bug fix for scenarios in which emissions are equal in the BAU and Optimized cases (e.g. system is sized but not dispatched). These scenarios and off grid runs now report "None" for the output "breakeven_cost_of_emissions_reduction_us_dollars_per_tCO2" 

Div by Zero error. 

These scenarios and off grid runs now report "None" for the output "breakeven_cost_of_emissions_reduction_us_dollars_per_tCO2" 

No.


Bill-Becker:
@adfarth Looks good! Do you know if we'll need to update the UI to handle None/null in the results in this case?

adfarth:
> @adfarth Looks good! Do you know if we'll need to update the UI to handle None/null in the results in this case?

I do not think so (other cases with None outputs show up as N/A in the outputs currently) but we can check with Rob/Nick.

Bill-Becker (approved):

PR #289, opened 2022-01-10
Fix double-counting CHP prod-based O&M cost

Bill-Becker:
Major CHP bug hot fix: CHP variable (production-based) O&M is being double-counted (twice as high as it should be). This change fixes that, and makes the reopt_model.jl code more concise.


adfarth:
Looks good!

adfarth (approved):

PR #290, opened 2022-01-11
CHP O&M bug fix and DivByZero error fix

Bill-Becker:
See https://github.com/NREL/REopt_Lite_API/pull/289 on hot fix for CHP O&M double-counting.

See https://github.com/NREL/REopt_Lite_API/pull/287 for divide by zero error fix when techs are sized but not dispatched.

REVERTED this celery bump because there was an error finding the newer version of celery: See https://github.com/NREL/REopt_Lite_API/pull/288 for Celery bump due to security vulnerability fix.

### DIFF
--- a/ghpghx/models.py
+++ b/ghpghx/models.py
@@ -97,8 +97,8 @@ class GHPGHXInputs(models.Model):
                             "5B": 1.177,
                             "6A": 0.977,
                             "6B": 0.981,
-                            "7A": 1.034,
-                            "7B": 1.508
+                            "7": 1.034,
+                            "8": 1.508
                             }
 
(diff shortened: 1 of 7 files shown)


Bill-Becker:
@adfarth The celery bump did not work, so I reverted that change and this PR should be ready to go. Thanks!

NLaws:
> The celery bump did not work

@Bill-Becker looks like we need to upgrade the Python version to use the latest celery release. https://docs.celeryproject.org/en/stable/getting-started/introduction.html

I will do this. Note that developers will need to rebuild their docker images to get the latest python version and packages.

adfarth (approved):

PR #293, opened 2022-01-20
update all python requirments (python 3.8)

NLaws:


reo/tests/test_chp.py
```diff
@@ -34,7 +34,7 @@
 from django.test import TestCase
 from reo.models import ModelManager
 from reo.utilities import check_common_outputs
-import numpy as np
+from numpy_financial import npv
```
zolanaj, line 37: We import the IRR calculation as follows: 
`from numpy_financial import irr as IRR`

I like using the capital notation for imported functions from other libraries, but either way, we should be consistent.

zolanaj (approved):
Thanks for this @nlaws! I have one minor comment, and I have bandwidth to execute the change if you want.

reo/tests/test_chp.py
NLaws: Makes sense to me to use the capitalization for `NPV` as well. (I capitalized `IRR` because the value is stored in `irr` - so watch out for name conflicts.) Yes please implement!

PR #294, opened 2022-01-21
update nested_inputs so off-grid defaults do not incorrectly appear f…

adfarth:
Previously, the off-grid defaults for sr_required_pct and min_load_met_pct would appear in the inputs nested dict returned from posting to the API, even for on-grid runs. However, these defaults do not get applied to on-grid runs, so this is likely to mislead users. Now, the default for min_load_met will show as 1.0 and sr_required (for load and PV) will show as 0 for on-grid runs. The applicability of these inputs to off-grid is also now specified in the descriptions.

Adds PV | `sr_required_pct` and LoadProfile | `sr_required_pct`, and LoadProfile | `min_load_met_pct` to the off_grid_defaults nested dict in nested_inputs.py. This allows for distinct defaults to be supplied for off- and on-grid runs. 

Currently, if a user does an on-grid analysis, they will see off-grid spinning reserve and min_load_met defaults in the inputs key of the returned nested dictionary. This is likely to be confusing, as the defaults don't actually get applied in on-grid runs. 

The default for min_load_met will show as 1.0 and sr_required (for load and PV) will show as 0 for on-grid runs. The applicability of these inputs to off-grid is also now specified in the descriptions.

No.


Bill-Becker:
@adfarth Sorry for the delay, this fell off my radar after I first started looking at it!

Bill-Becker (approved):

PR #313, opened 2022-03-26
Outage timestep corrections

hdunham:

bug fix

When normal and critical load profiles are spliced together for the BAU problem (and zeroed out after the time steps that BAU techs can sustain), missing minus ones in indexing is causing BAU problem to be infeasible when existing techs can sustain part but not all of outage.

The spliced load passed to the BAU problem contains the critical load in the outage time steps up to t (where t is the last consecutive time step that the existing techs can meet load) and zeros in the outage time steps after that.

no


NLaws (changes requested):
Please update the test re: my inline comment.

reo/tests/test_critical_load_bau.py
```diff
@@ -46,6 +46,39 @@ def setUp(self):
...
+        d = ModelManager.make_response(run_uuid=run_uuid)
+        c = nested_to_flat(d['outputs'])
+        c['resilience_check_flag'] = d['outputs']['Scenario']['Site']['LoadProfile']['resilience_check_flag']
+        c['bau_sustained_time_steps'] = d['outputs']['Scenario']['Site']['LoadProfile']['bau_sustained_time_steps']
+
+        d_expected = dict()
+        d_expected['status'] = 'optimal'
+        d_expected['year_one_energy_cost_bau'] = 1458525.98
+        d_expected['resilience_check_flag'] = False
+        d_expected['bau_sustained_time_steps'] = 22
```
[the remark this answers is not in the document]
hdunham: In order to create a test as quickly as possible, I just used the scenario from the user that was running into the bug. But I've updated the test now to something where we can manually calculate expected values.

reo/tests/test_critical_load_bau.py
```diff
@@ -58,20 +58,42 @@ def test_critical_load_bau_can_sustain_part_outage(self):
         nested_data = json.load(open(test_post, 'rb'))
 
         resp = self.get_response(data=nested_data)
-        self.assertHttpCreated(resp)
+        # self.assertHttpCreated(resp)
```
NLaws, line 61: Why is this line commented out?

NLaws (changes requested):
Please add a description to the doc string of `test_critical_load_bau_can_sustain_part_outage` of what is being tested so that we do not have to read the test line by line to understand what it is doing.  Can you please describe the blocks of calculations with comments? Is it only testing BAU values? (If so when we put it in REopt.jl we could run just a BAU scenario). It looks like `pwf_om` is calculated but not used.

adfarth (approved):
@hdunham @NLaws I think this is good to go with just the one unaddressed comment from Nick

reo/tests/test_critical_load_bau.py
hdunham: Commented out during testing and should have reverted

PR #326, opened 2022-06-08
Pv prod fix

adfarth:
- [x] CHANGELOG.md is updated
- [N/A ] Docs have been added / updated (for bug fixes / features) --> @NLaws do you think we should update the user manual for this? 

- In `src/pvwatts.py`, Updated lat-long coordinates if-statement used to determine whether the "nsrdb" dataset should be used to determine the PV prod factor. Accounts for recent updates to NSRDB data used by PVWatts (v6) 
- Avoids overwriting user-entered PV azimuth (other than 180) for ground-mount systems in southern hemisphere
- Updates default azimuth to 0 for southern latitudes for all PV types (rather than just for ground-mount)

- Incorrect bounds of NSRDB dataset are applied in PVWatts API call, meaning we are using lower quality PV resource data for some international sites
- Default azimuth remains at 180 for rooftop systems in southern hem
- Overwrites user-entered PV azimuth for ground-mount systems in southern hemisphere

See changes

Yes? More accurate PV data will be used for international evaluations. 

### Other information:
These are the bounds that are now applied to the NSRDB query: 
image


reo/src/pvwatts.py
```diff
@@ -134,8 +134,8 @@ def __init__(self,
             # if the site is in the southern hemisphere, and no tilt has been specified,
             # then set the tilt to the positive latitude value and change the azimuth to zero
             if self.latitude < 0:
```
NLaws, line 136: do we still need this logic in pvwatts.py (the whole if/else block)? It looks like it is taken care of in the PV Tech class

NLaws (changes requested):
A couple comments to address but looks good!

Can we combine your new test with this old one: 
https://github.com/NREL/REopt_API/blob/master/reo/tests/test_southern_hemisphere_latitude.py
?
Combining the two would make one less API post/optimization in the test suite.

reo/tests/test_indonesia_pv.py
```diff
@@ -0,0 +1,101 @@
...
+        r = json.loads(resp.content)
+        run_uuid = r.get('run_uuid')
+        d = ModelManager.make_response(run_uuid=run_uuid)
+        messages = d['messages']
+
+        a = d["outputs"]["Scenario"]["Site"]["PV"]
+        pv_list = a['year_one_to_battery_series_kw'], a['year_one_to_load_series_kw'], a['year_one_to_grid_series_kw'], a['year_one_curtailed_production_series_kw']
+        tot_pv = np.sum(pv_list, axis=0)
+        total_pv = sum(tot_pv)
+        print(total_pv)
```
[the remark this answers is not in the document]
adfarth: whoops sorry! Removed in https://github.com/NREL/REopt_API/pull/326/commits/b7cf74ea5a45e73085ada67e8385100e72621489

adfarth:
> A couple comments to address but looks good!
> 
> Can we combine your new test with this old one: https://github.com/NREL/REopt_API/blob/master/reo/tests/test_southern_hemisphere_latitude.py ? Combining the two would make one less API post/optimization in the test suite.

Combined these tests here! https://github.com/NREL/REopt_API/pull/326/commits/9eb58aa1915227b615b1849de2d0bf6783ec5807

reo/src/pvwatts.py
adfarth: No I think this was duplicative. Addressed in https://github.com/NREL/REopt_API/pull/326/commits/c9e51e3646b1365d8301d6e54450ab60e7db7b64

NLaws (approved):
Please merge if tests pass. We can also merge this into master after testing the develop deploy and merge to staging/production thereafter.

PR #328, opened 2022-06-24
Add off-grid capability and tax suffixes

rathod-b:
Add off-grid capability to API v3, add name changes to include before/after tax in result keys to api output models.

- [x] CHANGELOG.md is updated

Feature: the ability to provide off-grid inputs via Job app and the ability to receive off-grid specific outputs from Job app
Feature: the ability to tell if a calculated monetary value is before or after taxes from the returned keys

Currently, if off-grid inputs were to be provided to REopt via the Job app, they would get ignored because there are no existing models to accept to this data. Furthermore, the input processing can also fail if certain as-is required keys are omitted.

It is not possible to tell if a returned monetary value is before or after tax without looking at documentation.

Now, the user is able to provide off-grid scenarios to REopt via the Job app. Certain required inputs are now conditionally validated based on provided scenario.

It is now possible to tell if a returned monetary value is before or after tax without looking at documentation.

Yes, users' result processing scripts might break due to changes in keys names.


job/models.py
```diff
@@ -216,19 +218,27 @@ class TIME_STEP_CHOICES(models.IntegerChoices):
...
         help_text=("If True, then a small incentive to keep the battery's state of charge high is added to the "
                    "objective of the optimization.")
     )
+
     run_bau = models.BooleanField(
-        default=True,
         blank=True,
+        null=True,
         help_text=("If True then the Business-As-Usual scenario is also solved to provide additional outputs such as "
                    "the LCC and BAU costs.")
```
adfarth, line 233: Should this be "such as the NPV and BAU costs" ?

job/src/run_jump_model.py
```diff
@@ -81,6 +81,8 @@ def run_jump_model(run_uuid):
     name = 'run_jump_model'
     data = get_input_dict_from_run_uuid(run_uuid)
     user_uuid = data.get('user_uuid')
+    
+    data.pop('user_uuid',None) # Remove user uuid from inputs dict to avoid downstream errors
```
adfarth, line 85: @rathod-b was this something you discussed with Nick? Are there any downsides to doing this or potential issues?

job/test/test_job_endpoint.py
```diff
@@ -108,3 +108,99 @@ def test_pv_and_battery_scenario(self):
         self.assertAlmostEqual(results["ElectricStorage"]["size_kw"], 55.9, places=1)
         self.assertAlmostEqual(results["ElectricStorage"]["size_kwh"], 78.9, places=1)
 
+    def test_off_grid_defaults(self):
```
hdunham, line 111: Do these tests need to be in the API? Our thinking was to test all optimization functionality in the Julia package. Then in the API only test what's in the API repo, so making sure validation and the endpoint work.

job/validators.py
```diff
@@ -185,11 +185,13 @@ def clean_fields(self):
     def clean(self):
         """
         Run all models' clean methods
+        Run ElectricTariff clean method in cross-clean
```
hdunham, line 188: I think we should stick to the paradigm where clean() calls all the models' self-contained clean functions. Anything going into cross-clean should be named as such, take the existing cross_clean_pv() as an example.

hdunham: Actually maybe it's not important enough to spend time on right now. But, if we're only cleaning ElectricTariff if off_grid_flag is True, shouldn't we be doing the same thing with ElectricUtility?

job/test/test_job_endpoint.py
rathod-b: These tests are validating that API functionality (properly saving input JSONs, properly setting defaults given off-grid flag, and properly returning off-grid related fields in response JSON) is working and we are able to send and receive off-grid related fields from Julia. The `test_off_grid_defaults` test makes sure that the API is setting correct default field values conditional on off-grid flag. It also checks that the response JSON actually contains `off-grid` related fields. The second `override` test checks that a user is able to override the defaults set by previous test, which is a functionality of the API validator. These tests do hit the Julia API and run an optimization, but I am using very small input scenarios here because the goal isn't to validate REopt functionality, but to make sure the API is working as expected.

job/validators.py
rathod-b: I can try to find a way where clean() doesnt break if electric tariff is missing from inputs and off-grid is true. And we can create a new function to cross-clean electric tariff.

rathod-b: Updated per https://github.com/NREL/REopt_API/pull/328/commits/2c3e977834c2671ce76fcfdd32faed8dbe91d7f6

job/models.py
rathod-b: Updated in https://github.com/NREL/REopt_API/pull/328/commits/2c3e977834c2671ce76fcfdd32faed8dbe91d7f6

CHANGELOG.md
```diff
@@ -26,6 +26,49 @@ Classify the change according to the following categories:
...
+    - Added **offgrid_annual_oper_res_required_series_kwh**
+    - Added **offgrid_annual_oper_res_provided_series_kwh**
+    - Added **offgrid_load_met_series_kw**
+- **ElectricTariffInputs**
+    - Changed all instances of `coincident_peak_load_active_timesteps` to `coincident_peak_load_active_time_steps`
+- **ElectricTariffOutputs**
+    - Changed field names to add suffixes denoting `before_tax` or `after_tax` values
+- **ElectricTariffInputs**
+    - Changed validation of this model to be conditional on **Settings.off_grid_flag** being False
+`job/run_jump_model.py` - Remove `run_uuid` key from input dictionary before running REopt to avoid downstream errors from Julia
```
Bill-Becker, line 66: you removed `user_uuid`, right?

Bill-Becker (approved):
Looks good! I have just a couple minor comments, with the one about trying to avoid running the optimization model for the tests as just an idea. I'm going to approve.

CHANGELOG.md
rathod-b: Correct, it was causing errors from some downstream validations in REopt.jl

job/test/test_job_endpoint.py
rathod-b: I actually did not see that file at all. I have moved the off-grid default and off-grid default override tests into `test_validator.py` file. Should we have a way to test the output data models (i.e. that REopt response JSON includes the fields we think it should include)?

Bill-Becker: Yeah I'm not quite sure how to properly test the outputs without running the REopt model. There are ways to reduce the REopt model run time to speed up the tests, like setting the optimality tolerance higher, fixing sizes, reducing techs/loads, etc. I think it's fine if we have to run REopt.jl for some of the tests, but we just want to avoid it where we can.

Bill-Becker (approved):

job/validators.py
rathod-b: Tested to make sure that providing ElectricUtility does not break off_grid scenarios

PR #340, opened 2022-07-19
add capability to job/ app (API v3 using REopt.jl backend) to handle CHP

hdunham:
- [x] CHANGELOG.md is updated

Add capability to job/ app (API v3 using REopt.jl backend) to handle CHP technology


Bill-Becker:
@zolanaj When I POST a CHP run with a `prime_mover`, I'm getting an inputs error that it's expecting a few other inputs which should be assigned based on `prime_mover`.

"input_errors": {"CHP": {"installed_cost_per_kw": ["This field cannot be blank."], "tech_sizes_for_cost_curve": ["This field cannot be blank."], "unavailability_periods": ["This field cannot be blank."]}

Bill-Becker:
Looks like it's because those model fields don't have "blank=True" - just want to check if that was intentionally omitted?

Relatedly, if the user does NOT put in `prime_mover`, or they want use the same `installed_cost_per_kw`  for all sizes, we actually don't require a list for `installed_cost_per_kw` and `tech_sizes_for_cost_curve` because there should be an option to just put in a single value (scalar) for `installed_cost_per_kw` with no associated `tech_sizes_for_cost_curve` (it then assumes the `installed_cost_per_kw` is applied to any/all sizes).

zolanaj:
> @zolanaj When I POST a CHP run with a `prime_mover`, I'm getting an inputs error that it's expecting a few other inputs which should be assigned based on `prime_mover`.
> 
> "input_errors": {"CHP": {"installed_cost_per_kw": ["This field cannot be blank."], "tech_sizes_for_cost_curve": ["This field cannot be blank."], "unavailability_periods": ["This field cannot be blank."]}





> Looks like it's because those model fields don't have "blank=True" - just want to check if that was intentionally omitted?
> 
> Relatedly, if the user does NOT put in `prime_mover`, or they want use the same `installed_cost_per_kw` for all sizes, we actually don't require a list for `installed_cost_per_kw` and `tech_sizes_for_cost_curve` because there should be an option to just put in a single value (scalar) for `installed_cost_per_kw` with no associated `tech_sizes_for_cost_curve` (it then assumes the `installed_cost_per_kw` is applied to any/all sizes).

Thanks for catching the cause!  The fix that enables blank values for prime_mover notes is in https://github.com/NREL/REopt_API/pull/340/commits/f446966cbf2c456cde5656ffef32ad0a5350ce10 and the allowance for installed_cost_per_kw as a float and tech_sizes_for_cost_curve to be blank is fixed in https://github.com/NREL/REopt_API/pull/340/commits/1035b8a20b2d714c105c322c3001edca57a46fca and tested now in https://github.com/NREL/REopt_API/pull/340/commits/1bb59129795eee5a30a1b2554e730bb25ae57e68.  There are a couple additional things I caught in here.

Could you let me know if the post you mentioned above works correctly now, and if not, could you send me your post so that I can ensure the new test covers that contingency?

Bill-Becker:
@zolanaj I'm now getting an error when validators.py tries to `model.save()` on line 539. Have you been able to POST a job with CHP?

job/models.py
```diff
@@ -3350,6 +3331,689 @@ class GeneratorOutputs(BaseModel, models.Model):
...
+        elif len(self.dict.get("tech_sizes_for_cost_curve")) != len(self.dict.get("installed_cost_per_kw")):
+            error_messages["required inputs"] = \
+                "To model CHP cost curve, you must provide `chp.tech_sizes_for_cost_curve` vector of equal length to `chp.installed_cost_per_kw`"
+        
+        if not at_least_one_set(self.dict, self.possible_sets):
+            error_messages["required inputs"] = \
+                "Must provide at least one set of valid inputs from {}.".format(self.possible_sets)
+        else:  #use defaults if CHP prime mover given 
+            if self.dict.get("prime_mover") not in [None, ""]:
+                boiler_type = "steam" if self.prime_mover == "combustion_turbine" else "hot_water"
```
Bill-Becker, line 3875: @zolanaj I'm realizing now that this assignment of CHP defaults will have to go in validators.py because "steam" or "hot_water" could be a user-input for ExistingBoiler which would change this selection for thermal_efficiency. We also may in the future do the logic of size_class based on the average heating load, which would require the cross_clean functionality. Thoughts or downsides of moving this there?

Bill-Becker: I also just realized this same logic for assigning CHP defaults is replicated in REopt.jl. Thoughts on the need to replicate this type of conditional logic in the API? The obvious benefit is that the django models get assigned the default value, whereas if we just relied on REopt.jl then the api_response would not include those assigned values in the inputs. I'm planning to integrate the API's chp_defaults view/endpoint logic into REopt.jl so that the `size_class` gets chosen based on the average heating load, so that would also have to be replicated in the API validators.py. It starts to get a bit complicated to be replicating in two places.

Bill-Becker: FYI I'm going to work on creating a Julia function that will contain all of the API's `/chp_defaults` view logic and assign default values. So that will be used on the Julia-side, and I'll expose that function in `http.jl` as a Julia endpoint for validators to call so that we don't have to duplicate the logic. I think it shouldn't be a problem to call an `http.jl` endpoint in validators, right?

zolanaj: @Bill-Becker sorry this took a while to respond.  I'm in favor of simplifying the API and doing the work in Julia as much as possible the way we do for financials.  If I'm understanding your method of exposing the Julia-side assignment of default values in `http.jl`, Julia would return the default values that would then update the inputs before `optimize!` is called to generate what becomes the outputs later - is that right?  If so, I think you found an elegant way of avoiding duplicate input handling and I support it.

Bill-Becker:
@zolanaj Still waiting on the results of the test suite, but this should be good to re-review for you - I'll debug any tests that are still failing. I made pretty sweeping changes of your implementation of CHP defaults and validation - sorry about that, I didn't realize the most efficient way forward was to use REopt.jl's REoptInputs.s (Scenario) to assign the defaults back to models.py CHPInputs fields until after I was way into it. I made equivalent tests of the tests that you made for CHP defaults and validation, consistent with the revised processing.

Note, this branch is currently using the simulated-load branch of REopt.jl because it has integrated some of those features. We'll have to merge that stuff into REopt.jl master before merging this.

Bill-Becker (approved):
I updated to the latest published version of REopt now that the simulated-load branch was merged in.

zolanaj (approved):
Thanks for all the work on this @Bill-Becker!  This looks ready to merge to me.  In addition to handling some merge conflicts with devleop in `Manifest.toml`, The only things I added were a couple small tests to ensure that bad and missing inputs in the simulated_load function were returning the errors we expect.  Once the tests pass, I think this is good to go.

PR #342, opened 2022-07-26
/stable points to v2 reo/job

Bill-Becker:
- [x] CHANGELOG.md is updated

**I deployed `develop` to the development server and tested that we are getting the new defaults**

- URL change for using `v2` for `/stable`
- Bunch of `/job` additions for v3 development, see #328 
- an outage time step bug fix, see #313 
- dependency security patches, see #333 

- `/stable` URL for the `/job` endpoint uses Job() API which is the `v1` version
- see #328  #313  #333 

- `/stable` now uses the `Job2()` API which is the v2 version of `reo`
- see #328  #313  #333 

Yes, for `/stable/job` endpoint, but this fixes what was broken because we had intended `/stable` to be `v2`

### Other information:
- I'm going to display a banner on the webtool that we just made these changes to actually use the new defaults
- I'm gong to make an announcement on the User Forum announcing the change for /stable to v2 /job


hdunham (approved):
Looks good. I edited the description to mention the other bug fixes that are included in this PR besides pointing stable to the correct version. I also added a couple other minor commits to make sure the correct version meta data gets returned.

PR #343, opened 2022-08-01
add offgrid wind to job/ app, update generator and PV defaults

adfarth:
- [x] CHANGELOG.md is updated
- [ N/A] Docs have been added / updated (for bug fixes / features)

##### Added 
 - `job/` endpoint: Add inputs and validation to model off-grid wind 
 ##### Changed
 - `job/models.py`: remove Generator `fuel_slope_gal_per_kwh` and `fuel_intercept_gal_per_hr` defaults based on size, keep defaults independent of size 
 - `job/validators.py`: Align PV tilt and aziumth defaults with API v2 behavior, based on location and PV type

- Cannot model wind when `off_grid_flag=True`
- Generator fuel intercept and slope defaults set dependent upon size 
- PV tilt and azimuth defaults do not reflect API v2 behavior

See above

The PV and generator default changes are breaking as compared to previous v3 (job/ app) behavior, but not as compared to v2.


adfarth:
@rathod-b  do we need to use a dummy default tilt for PV in validators.py [here](https://github.com/NREL/REopt_API/blob/4b1a154ab82380f39ee16d12b6047e279f67a944/job/validators.py#L212)?

adfarth:
@rathod-b  do we need to provide [this message](https://github.com/NREL/REopt_API/blob/42abce2cf51744e37017197d2a5fe17ead7cbec2/job/validators.py#L151) in `msg_dict["info"]` every time off_grid_flag is true?: 
```if self.models["Settings"].off_grid_flag==True:
            if "ElectricTariff" in self.models.keys():
                msg_dict["ignored inputs"] = ("ElectricTariff inputs are not applicable when off_grid_flag is true, and will be ignored. "
                                "Provided ElectricTariff can be removed from inputs")
            msg_dict["info"] = ("When off_grid_flag is true, only PV, ElectricStorage, Generator technologies can be modeled.")```

Could we just show it if a disallowed key is supplied, as we do in the Julia Pkg [here](https://github.com/NREL/REopt.jl/blob/8d7beb8a916d1cdf8571c268d44a6ef86dc07a7b/src/core/scenario.jl#L88)?

File: job/validators.py [hunk 1]
```diff
@@ -209,42 +209,42 @@ def cross_clean(self):
         PV tilt set to latitude if not provided and prod_factor_series validated
         """
         def cross_clean_pv(pvmodel):
-            if pvmodel.__getattribute__("tilt") == 0.537:  # 0.537 is a dummy number, default tilt
-                pvmodel.__setattr__("tilt", self.models["Site"].__getattribute__("latitude"))
+            if pvmodel.__getattribute__("tilt") == 0.537:  # 0.537 is a dummy number, default tilt # TODO: change to None?
```
adfarth, line 212: @rathod-b The PV tilt defaults should use the same logic as the Julia pkg (based on array type).

From the Julia pkg: `tilt::Real= array_type == 1 ? 10 : abs(latitude), # tilt = 10 deg for rooftop systems, abs(lat) for ground-mount` 

Could you update this?

job/validators.py
```diff
@@ -209,42 +209,43 @@ def cross_clean(self):
         PV tilt set to latitude if not provided and prod_factor_series validated
         """
         def cross_clean_pv(pvmodel):
-            if pvmodel.__getattribute__("tilt") == 0.537:  # 0.537 is a dummy number, default tilt
-                pvmodel.__setattr__("tilt", self.models["Site"].__getattribute__("latitude"))
+            if pvmodel.__getattribute__("tilt") == 0.537:  # 0.537 is a dummy number, default tilt # TODO: change to None? 
+                pvmodel.__setattr__("tilt", self.models["Site"].__getattribute__("latitude")) # TODO differentiate default tilt based on array_type
+            # TODO add cross_clean for azimuth
```
adfarth, line 214: @rathod-b  We should also follow the Julia Pkg default behavior for PV azimuth:
`azimuth::Real = latitude≥0 ? 180 : 0, # set azimuth to zero for southern hemisphere` 

Could you make this update as well?

job/validators.py
```diff
@@ -209,42 +209,43 @@ def cross_clean(self):
...
             if pvmodel.__getattribute__("max_kw") > 0:
                 if len(pvmodel.__getattribute__("prod_factor_series")) > 0:
                     self.clean_time_series("PV", "prod_factor_series")
+            
+            if self.models["PV"].__getattribute__("can_net_meter") == None:
```
adfarth, line 219: @rathod-b this cross_clean appropriately set the default values if the keys (e.g. `can_net_meter`) are **not provided**. However, we also need to override user inputs if they, for example, set `can_net_meter=True` and `off_grid_flag=True`. We currently do this in the Julia Pkg [here](https://github.com/NREL/REopt.jl/blob/8d7beb8a916d1cdf8571c268d44a6ef86dc07a7b/src/core/pv.jl#L181) in pv.jl. So, if the user did incorrectly set one of these keys, it would get updated in Julia but the input saved in the database would not align with the input used in Julia. 

@Bill-Becker or @hdunham  I'm guessing we might have more instances of this. Do you think we should include every instance of overriding of incorrect user inputs (again, such as [this](https://github.com/NREL/REopt.jl/blob/8d7beb8a916d1cdf8571c268d44a6ef86dc07a7b/src/core/pv.jl#L181)) in the API validators.py as well as in the Julia Pkg?

job/validators.py
```diff
@@ -277,6 +278,12 @@ def cross_clean_pv(pvmodel):
                                                   self.models["Site"].__getattribute__("longitude")):
                         self.add_validation_error("Wind", "Site",
                               "latitude/longitude not in the WindToolkit database. Cannot retrieve wind resource data.")
+            
+            if self.models["Wind"].__getattribute__("operating_reserve_required_pct") == None:
```
adfarth, line 282: @rathod-b similar comment to above, we may also want to override user inputs if they provide operating reserve for wind and off_grid_flag is False.

adfarth:
@rathod-b This is mostly for my own edification, but do we (and why?) need `null=True` in models.py if we are also providing a default value, such as for the `offgrid_other_capital_costs` [here](https://github.com/NREL/REopt_API/blob/4b1a154ab82380f39ee16d12b6047e279f67a944/job/models.py#L427)?

job/validators.py
```diff
@@ -357,7 +364,7 @@ def cross_clean_pv(pvmodel):
             if self.models["Settings"].off_grid_flag==False:
                 self.models["ElectricLoad"].critical_load_pct = 0.5
             else:
-                self.models["ElectricLoad"].critical_load_pct = 1.0
+                self.models["ElectricLoad"].critical_load_pct = 1.0 # TODO: this always needs to be true for off-grid
```
adfarth, line 367: @rathod-b another instance of needing to override user inputs. Critical_load_pct must always = 1 when off_grid_flag is true. See the validation we do in REopt.jl [here](https://github.com/NREL/REopt.jl/blob/8d7beb8a916d1cdf8571c268d44a6ef86dc07a7b/src/core/electric_load.jl#L125).

job/validators.py
```diff
@@ -390,7 +397,7 @@ def cross_clean_pv(pvmodel):
 
             if self.models["Generator"].__getattribute__("replacement_year") == None:
                 if self.models["Settings"].off_grid_flag==False:
-                    self.models["Generator"].replacement_year = 25
+                    self.models["Generator"].replacement_year = 25 # TODO: set to analysis_years
```
adfarth, line 400: @rathod-b  this should be set to analysis_years when off_grid_flag is False, instead of 25.

job/validators.py
```diff
@@ -390,13 +397,13 @@ def cross_clean_pv(pvmodel):
...
                 else:
                     self.models["Generator"].replacement_year = 10
 
             if self.models["Generator"].__getattribute__("replace_cost_per_kw") == None:
                 if self.models["Settings"].off_grid_flag==False:
-                    self.models["Generator"].replace_cost_per_kw = 410.0
+                    self.models["Generator"].replace_cost_per_kw = 410.0 # TODO: set to 0
```
adfarth, line 406: @rathod-b this should be 0 to align with Julia Pkg default (replacement costs not considered by default for on-grid scenarios)

job/validators.py
rathod-b: Sure, I can update this.

rathod-b:
> @rathod-b do we need to use a dummy default tilt for PV in validators.py [here](https://github.com/NREL/REopt_API/blob/4b1a154ab82380f39ee16d12b6047e279f67a944/job/validators.py#L212)? Could we not set null=True in models.py and update in validators as we do for other inputs (e.g. operating reserves required)?

We can do that. I will make this update.

rathod-b:
> @rathod-b do we need to provide [this message](https://github.com/NREL/REopt_API/blob/42abce2cf51744e37017197d2a5fe17ead7cbec2/job/validators.py#L151) in `msg_dict["info"]` every time off_grid_flag is true?:
> 
> ```
>             if "ElectricTariff" in self.models.keys():
>                 msg_dict["ignored inputs"] = ("ElectricTariff inputs are not applicable when off_grid_flag is true, and will be ignored. "
>                                 "Provided ElectricTariff can be removed from inputs")
>             msg_dict["info"] = ("When off_grid_flag is true, only PV, ElectricStorage, Generator technologies can be modeled.")```
> 
> Could we just show it if a disallowed key is supplied, as we do in the Julia Pkg [here](https://github.com/NREL/REopt.jl/blob/8d7beb8a916d1cdf8571c268d44a6ef86dc07a7b/src/core/scenario.jl#L88)? 
> ```

I wanted to provide this message for informational purposes, similar to how we provide message on year being switched to 2017 if DOE reference profiles are provided for ElectricLoad. We can alter this to be more meaningful or remove it if necessary.

CHANGELOG.md
```diff
@@ -54,26 +54,80 @@ Classify the change according to the following categories:
     - Added **replacements_future_cost_after_tax**
     - Added **replacements_present_cost_after_tax**
     - Added **offgrid_microgrid_lcoe_dollars_per_kwh**
-    - Changed **lifecycle_capital_costs_plus_om** and **lifecycle_om_costs_bau** field names to include before/after tax
+    - Changed **lifecycle_capital_costs_plus_om** field name to **lifecycle_capital_costs_plus_om_after_tax**
+    - Changed **lifecycle_om_costs_bau** field name to **lifecycle_om_costs_before_tax_bau**
+- **ElectricLoadInputs**
+    - Removed default value for **critical_load_met_pct**. If user does not provide this value, it is defaulted depending on **Settings -> off_grid_flag**
+    - Added **operating_reserve_required_pct**
```
hdunham, line 61: this was added to WindInputs not ElectricLoadInputs

reo/nested_inputs.py
```diff
@@ -286,7 +286,7 @@ def list_of_dict(input):
         "type": "float",
         "min": 0.0,
         "max": 1.0e6,
-        "description": "Land area in acres available for PV panel siting"
+        "description": "Land area in acres available for PV panel and wind turbine siting"
```
hdunham, line 289: wind is not currently being constrained by land_acres

job/validators.py
```diff
@@ -209,42 +209,43 @@ def cross_clean(self):
         PV tilt set to latitude if not provided and prod_factor_series validated
         """
         def cross_clean_pv(pvmodel):
-            if pvmodel.__getattribute__("tilt") == 0.537:  # 0.537 is a dummy number, default tilt
-                pvmodel.__setattr__("tilt", self.models["Site"].__getattribute__("latitude"))
+            if pvmodel.__getattribute__("tilt") == None:  # 0.537 is a dummy number, default tilt # TODO: change to None?
```
hdunham, line 212: delete comment/todo

job/validators.py
```diff
@@ -209,42 +209,43 @@ def cross_clean(self):
...
+                else:
+                    pvmodel.__setattr__("tilt", abs(self.models["Site"].__getattribute__("latitude"))) # TODO differentiate default tilt based on array_type
+            
+            if pvmodel.__getattribute__("azimuth") == None:
+                if self.models["Site"].__getattribute__("latitude") >= 0:
+                    pvmodel.__setattr__("azimuth", 180)
+                else:
+                    pvmodel.__setattr__("azimuth", 0.0)
+            
+            # TODO add cross_clean for azimuth
```
hdunham, line 224: delete todo

job/validators.py
```diff
@@ -209,42 +209,43 @@ def cross_clean(self):
...
                 if len(pvmodel.__getattribute__("prod_factor_series")) > 0:
                     self.clean_time_series("PV", "prod_factor_series")
+            
+            if self.models["Settings"].off_grid_flag==True:
+                self.models["PV"].can_net_meter = False
+                self.models["PV"].can_wholesale = False
+                self.models["PV"].can_export_beyond_nem_limit = False
+                self.models["PV"].operating_reserve_required_pct = 0.25
+            else:
+                if self.models["PV"].__getattribute__("can_net_meter") == None:
```
hdunham, line 235: instead of these 3 if statements, why not just make the defaults in models.py True, then only keep the above if off grid block not the else block?

hdunham: the help and inputs endpoints use those defaults in models.py so they won't show up in the web tool input boxes otherwise. Off grid isn't in the web tool right? otherwise we would need on and off grid versions of help.

job/validators.py
```diff
@@ -277,6 +278,20 @@ def cross_clean_pv(pvmodel):
                                                   self.models["Site"].__getattribute__("longitude")):
                         self.add_validation_error("Wind", "Site",
                               "latitude/longitude not in the WindToolkit database. Cannot retrieve wind resource data.")
+            
+            if self.models["Settings"].off_grid_flag==True:
```
hdunham, line 282: same comments as PV above

job/validators.py
```diff
@@ -352,12 +367,15 @@ def cross_clean_pv(pvmodel):
         ElectricLoad
         If user does not provide values, set defaults conditional on off-grid flag
         """
-
-        if self.models["ElectricLoad"].__getattribute__("critical_load_pct") == None:
-            if self.models["Settings"].off_grid_flag==False:
-                self.models["ElectricLoad"].critical_load_pct = 0.5
-            else:
+        '''
```
hdunham, line 370: Similar comment to above. If we're just overriding in the pff grid situation anyways, we can just make the Django default the on grid default (0.5) and delete the else block below

job/test/test_validator.py
```diff
@@ -167,6 +170,7 @@ def test_off_grid_defaults_overrides(self):
         validator.cross_clean()
         self.assertEquals(validator.is_valid, True)
 
-        self.assertAlmostEqual(validator.models["ElectricLoad"].critical_load_pct, 0.95)
+        self.assertAlmostEqual(validator.models["ElectricLoad"].critical_load_pct, 1.0)
+        self.assertAlmostEqual(validator.models["Wind"].operating_reserve_required_pct, 0.1)
```
hdunham, line 174: why is the wind operating reserve requirement always 0.1 regardless of what the user sets it to? What's the point of the input then?

hdunham:
I've finished reviewing and left some comments. The PR description also needs to be filled out.

job/validators.py
adfarth: @hdunham just to be certain, do you happen to know if there's any reason we need to use this dummy value of 0.537 as opposed to setting the default to None as we do with other inputs?

job/test/test_validator.py
rathod-b: Good point, we have a few inputs like this where we want to override user inputs or have that input be 0.0. We still want to have them as inputs so users can get the value used by REopt when they GET their responses.

reo/nested_inputs.py
adfarth: @hdunham I believe it is in v2, [here](https://github.com/NREL/REopt_API/blob/4b1a154ab82380f39ee16d12b6047e279f67a944/reo/src/data_manager.py#L1036). 

It is _not_ in v3/Julia Pkg, currently, but we've added an issue for this [here](https://github.com/NREL/REopt.jl/issues/108)

job/test/test_validator.py
adfarth: @rathod-b the wind `operating_reserve_required_pct` should **not** be overridden if `off_grid_flag=True`. The default should be **0.5** but a user should be able to provide another value. If on-grid, it **should** be forced to be 0.

job/validators.py
adfarth: @hdunham off-grid is in the web tool. You uncheck "Grid" to model off-grid.

job/validators.py
```diff
@@ -209,42 +209,36 @@ def cross_clean(self):
...
+            
             if pvmodel.__getattribute__("max_kw") > 0:
                 if len(pvmodel.__getattribute__("prod_factor_series")) > 0:
                     self.clean_time_series("PV", "prod_factor_series")
+            
+            if self.models["Settings"].off_grid_flag==True:
+                self.models["PV"].can_net_meter = False
+                self.models["PV"].can_wholesale = False
+                self.models["PV"].can_export_beyond_nem_limit = False
+                self.models["PV"].operating_reserve_required_pct = 0.25
```
adfarth, line 232: @rathod-b PV `operating_reserve_required_pct` should not be overwritten if `off_grid_flag=True`. The default should be 0.25 but the user should be allowed to provide another value. Please adjust the validation to reflect this.

job/validators.py
```diff
@@ -277,6 +271,14 @@ def cross_clean_pv(pvmodel):
                                                   self.models["Site"].__getattribute__("longitude")):
                         self.add_validation_error("Wind", "Site",
                               "latitude/longitude not in the WindToolkit database. Cannot retrieve wind resource data.")
+            
+            if self.models["Settings"].off_grid_flag==True:
+                self.models["Wind"].operating_reserve_required_pct = 0.10
```
adfarth, line 276: @rathod-b Load`operating_reserve_required_pct` should not be overwritten if `off_grid_flag=True`. The default should be 0.1 but the user should be allowed to provide another value. Please adjust the validation to reflect this.

job/validators.py
```diff
@@ -352,24 +354,17 @@ def cross_clean_pv(pvmodel):
...
+        If !off-grid scenario, if critical load pct is not specified in inputs, set it to 0.5. Otherwise, allow user specified pct.
+        '''
+        if self.models["Settings"].off_grid_flag==True:
                 self.models["ElectricLoad"].critical_load_pct = 1.0
-        
-        if self.models["ElectricLoad"].__getattribute__("operating_reserve_required_pct") == None:
-            if self.models["Settings"].off_grid_flag==False:
-                self.models["ElectricLoad"].operating_reserve_required_pct = 0.0
-            else:
                 self.models["ElectricLoad"].operating_reserve_required_pct = 0.1
```
adfarth, line 363: @rathod-b Load `operating_reserve_required_pct` should not be overwritten if `off_grid_flag=True`. The default should be 0.1 but the user should be allowed to provide another value. Please adjust the validation to reflect this.

job/validators.py
```diff
@@ -352,24 +354,17 @@ def cross_clean_pv(pvmodel):
...
-
-        if self.models["ElectricLoad"].__getattribute__("min_load_met_annual_pct") == None:
-            if self.models["Settings"].off_grid_flag==False:
-                self.models["ElectricLoad"].min_load_met_annual_pct = 1.0
-            else:
                 self.models["ElectricLoad"].min_load_met_annual_pct = 0.99999
```
adfarth, line 364: @rathod-b Similar to above comment, `min_load_met_annual_pct` should not get overwritten if `off_grid_flag=True`. Please change validators.py to reflect this.

job/validators.py
adfarth: removed comment in https://github.com/NREL/REopt_API/pull/343/commits/066d132881ab26f9c5d9270a3700244f37e23d3c

job/validators.py
adfarth: removed

CHANGELOG.md
adfarth: This was retroactively updated for the previous PR, in which `operating_reserve_required_pct` was in fact added to `ElectricLoadInputs`. I've updated the CHANGELOG for this PR to include `operating_reserve_required_pct` for WindInputs

job/validators.py
adfarth: updated

job/validators.py
adfarth: updated

hdunham:
> > @rathod-b This is mostly for my own edification, but do we (and why?) need `null=True` in models.py if we are also providing a default value, such as for the `offgrid_other_capital_costs` [here](https://github.com/NREL/REopt_API/blob/4b1a154ab82380f39ee16d12b6047e279f67a944/job/models.py#L427)?
> 
> Technically we can remove the `null=True` from fields like this because per my understanding, setting a `default` value means this field is not going to be null (unless the code fails). If we were to update the default value downstream in `cross_clean()` then we would remove the default here so the field can be set correctly in cross_clean (as we do with PV tilt now).

My understanding is also that null=True isn't necessary if a default is provided, though I haven't been able to find this said explicitly in the Django docs so had been holding off on the issue. Removing all unnecessary null=True is something that I've been keeping in mind as a low priority to-do for a while and was going to include in a general validator.py/models.py clean up after finishing more urgent capabilities.

rathod-b:
> > > @rathod-b This is mostly for my own edification, but do we (and why?) need `null=True` in models.py if we are also providing a default value, such as for the `offgrid_other_capital_costs` [here](https://github.com/NREL/REopt_API/blob/4b1a154ab82380f39ee16d12b6047e279f67a944/job/models.py#L427)?
> > 
> > 
> > Technically we can remove the `null=True` from fields like this because per my understanding, setting a `default` value means this field is not going to be null (unless the code fails). If we were to update the default value downstream in `cross_clean()` then we would remove the default here so the field can be set correctly in cross_clean (as we do with PV tilt now).
> 
> My understanding is also that null=True isn't necessary if a default is provided, though I haven't been able to find this said explicitly in the Django docs so had been holding off on the issue. Removing all unnecessary null=True is something that I've been keeping in mind as a low priority to-do for a while and was going to include in a general validator.py/models.py clean up after finishing more urgent capabilities.

Good idea, once we merge everything into API v3 over the next few weeks it would be nice to clean up the code and remove redundancies/make functions for repetitive validations in model cross_clean

job/validators.py
rathod-b: I have updated the PV field validations (thank you Hallie!) for `can_net_meter`, `can_export_beyond_nem_limit`, `can_wholesale`, so we default these to True in models.py but set them to False if off-grid is true.

Currently, `operating reserve required pct` is not defaulting to any value. We set this field to 0.25 if off-grid is true, and if off-grid is false, we override anything the user might have provided and set this value to 0.0 per [here](https://github.com/NREL/REopt.jl/blob/8d7beb8a916d1cdf8571c268d44a6ef86dc07a7b/src/core/pv.jl#L176) (defaulting it will allow user to provide an input value, which we dont want). I have also updated the help text for this field so users know this gets over-written regardless of what they provide (if they provide a value).

@adfarth we can know if user provided an input by checking if `operating_reserve_required_pct` is None. But it is my understanding that we dont want any ability to modify this field? It has to be 0.25 for off-grid and 0.0 for on grid? I am not seeing us enforcing this field to be 25% in julia though.

adfarth: @rathod-b see my comments below regarding `operating_reserve_required_pct` for `PV, Wind, and Load`. These should **not** get overwritten when `off_grid_flag=True`. They should default to their respective values (shown in Julia code) but the user can modify these inputs for off-grid scenarios.

job/test/test_validator.py
rathod-b: Understood, i was wondering if that was the case for `operating_reserve_required_pct` for PV and ElectricLoad models as well?

job/validators.py
rathod-b: Addressed in https://github.com/NREL/REopt_API/pull/343/commits/152ac98842eaaf16a7068f0412b34b0893f0886a

job/validators.py
rathod-b: Addressed in https://github.com/NREL/REopt_API/pull/343/commits/152ac98842eaaf16a7068f0412b34b0893f0886a

job/validators.py
rathod-b: Addressed in https://github.com/NREL/REopt_API/pull/343/commits/152ac98842eaaf16a7068f0412b34b0893f0886a

job/validators.py
rathod-b: Addressed in https://github.com/NREL/REopt_API/pull/343/commits/152ac98842eaaf16a7068f0412b34b0893f0886a

job/test/test_validator.py
rathod-b: Addressed in https://github.com/NREL/REopt_API/pull/343/commits/152ac98842eaaf16a7068f0412b34b0893f0886a

job/validators.py
rathod-b: Implemented in https://github.com/NREL/REopt_API/pull/343/commits/152ac98842eaaf16a7068f0412b34b0893f0886a

job/validators.py
rathod-b: Implemented in https://github.com/NREL/REopt_API/pull/343/commits/152ac98842eaaf16a7068f0412b34b0893f0886a

job/validators.py
rathod-b: Updated per prior comment on PV Inputs

reo/nested_inputs.py
hdunham: So then can we not edit this description until we actually make the change?

job/validators.py
hdunham: Yep I mostly agree with the situations you laid out @adfarth . Sometimes we need to know if the user didn't enter a value, but for the inputs where in one grid case we take the user input/default and the other grid case we override no matter what then we don't need to know. 
The one part I don't completely agree on is the second situation you listed, which applies to operating_reserve_required_pct. Technically, we don't need to know if the user provided a value. We could make the Django default equal to off-grid default and if offgrid=false then override it to 0. Essentially just the reverse of the first situation in your list. We might not want to do this because then there would be a mix of on- and off- grid defaults in the inputs endpoint. Or, maybe that's fine because the help text explains when an input is only applicable to off-grid. Just wanted to point out that there are two options to decide between, each with pros and cons.

hdunham: I also left a couple comments on help text in [https://github.com/NREL/REopt_API/commit/152ac98842eaaf16a7068f0412b34b0893f0886a](url)

hdunham: I also left a couple comments on help text in [152ac98](https://github.com/NREL/REopt_API/commit/152ac98842eaaf16a7068f0412b34b0893f0886a)

hdunham: When I said having on and off grid versions of help, I mean of the help/ endpoint (equivalent to inputs/ endpoint). That would mean creating a new function in views.py.

reo/nested_inputs.py
adfarth: @hdunham this change is in `reo/nested_inputs.py`, i.e. `v2`, and it is accurately describing v2 behavior. Am I missing something?

job/validators.py
adfarth: @hdunham I agree with the logic you've laid out! I have been thinking that we wouldn't want to mix on- and off-grid defaults in `models.py`, though. I think we should ask the broader team about this

rathod-b (approved):

reo/nested_inputs.py
hdunham: Nope it was me that was missing that this change was in reo/

job/validators.py
hdunham: I made an issue so we can remember to address this later as it's lower priority at the moment I think.

PR #350, opened 2022-08-05
Validator improvements

hdunham:
- [x] CHANGELOG.md is updated

Bug fixes

ElectricTariff key is never required, even for on-grid
ElectricUtility model not created if not provided in inputs of an on-grid analysis
ElectricUtility silently ignored if provided in inputs for an off-grid analysis

Require ElectricTariff key for on-grid
Create and save ElectricUtility model if not provided in inputs of an on-grid analysis, in order to have the default inputs
Warn that ignoring ElectricUtility if provided in inputs for an off-grid analysis
Add test to ensure errors returned when required keys are missing

no


zolanaj (approved):
@hdunham Thanks, this looks great!  I added the following tests to yours: 
- Assert that the starting post doesn't trigger any missing required input errors, with `post["Settings"]["off_grid_flag"]` set to True and False (might be overkill, but could be helpful if we add separate required inputs to either ongrid or offgrid in the future)
- When `post["Settings"]["off_grid_flag"] = True`, assert that `"ElectricTariff"` inputs do not show up in the validation errors, even if missing

If you agree with this, please go ahead and merge.  Thanks again!

PR #351, opened 2022-08-09
Update deployment code and dependencies

GUI:
- Update deployments to use Werf 1.2 which brings various improvements and simplifies some things.
- This upgrade to the underlying deployment image also brings a newer version of kubectl which I think is why some recent deployments have been failing against upgraded kubernetes clusters (since the old deployment code wasn't using a new enough version of kubectl tor rancher's cli).
- Store the deployed images in ECR so that we can also implement a new image cleanup job. This helps cleanup unused images from old deploys from the image repository so the storage needs don't grow indefinitely.
- Address some issues with ingress version and annotations needing to be upgraded for new versions of kubernetes. But note that we currently need slightly different approaches for dev and staging/prod due to different Kubernetes versions these clusters are running. But once staging/prod are on the same version of Kubernetes as dev, then this will need to be updated so they all use the same approach.

I think these changes will need to be merged into any branches you want to fix the deploys on. So not totally sure how you're branching strategy is working and whether you want this in `developer` or `master` or how that might propagate to other branches.


adfarth:
> 

@Bill-Becker, @GUI made changes in the `nova-analysis2` branch that address this undeploy issue. Nick mentioned he would pull those changes into here while making the resource limit changes as well.

GUI:
Yes, the environment variable issue should be fixed by this commit I pulled back into this branch: https://github.com/NREL/REopt_API/pull/351/commits/b6acb51d725f112b00a35f79e7db1389d289f046

I've also added configuration to setup resource requests/limits as we had talked about to try and help with potential memory spikes: https://github.com/NREL/REopt_API/pull/351/commits/e673ac7af091fc859a5c9097ad5e8f928f8963ef

Bill-Becker (approved):

PR #362, opened 2022-09-13
Pct name changes

adfarth:
- [x] CHANGELOG.md is updated
- [N/A ] Tests for the changes have been added (for bug fixes / features)

Input and output name changes for names containing "_pct", only in the `job/` app (v3). This is because these values are actually fractions (between 0-1) and should not be interpreted as percentages (0-100%). 

Names contain "pct" but are not truly percentages. 

The following name changes were made in the `job/` endpoint: 
 - Change "_pct" to "_rate_fraction" for input and output names containing "discount", "escalation", and "tax_pct" (financial terms)
 - Change "_pct" to "_fraction" for all other input and output names (e.g., "min_soc_", "min_turndown_")
 - Change **prod_factor_series** to **production_factor_series** 

Only for v3 (not currently used by users)

### Note!
The following name changes were also made in REopt.jl, but these names are not yet included in REopt_API. These changes will thus be reflected in a future API PR: 
- four (4) CHP input field names to spell out `electric` (from `elec`) and `efficiency` (from `effic`) for electric and thermal efficiencies


adfarth:
@Bill-Becker  we should re-run the tests and then merge this after the equivalent Julia Pkg changes are merged and registered (https://github.com/NREL/REopt.jl/pull/131) .

Bill-Becker (approved):
@adfarth looks good! Do you think our current job app tests actually test these kinds of name changes well enough (where the test would fail if names between REopt.jl and REopt_API were different)? I see there is a POST in test_job_endpoint.py that has some of these changed inputs.

adfarth:
> @adfarth looks good! Do you think our current job app tests actually test these kinds of name changes well enough (where the test would fail if names between REopt.jl and REopt_API were different)? I see there is a POST in test_job_endpoint.py that has some of these changed inputs.

Yes, input and output name discrepancies between the API and Julia pkg will get caught, even if not included in the job app posts. I believe this is because defaults of any inputs not provided in an API post will still get populated and then the Julia Pkg will send an error (and vice versa for results from Julia)

PR #379, opened 2022-11-01
Fix lookback

Bill-Becker:
This fixes the wrong-capitalization of lookback parameters expected from URDB, so that lookback works. 

The existing test_demand_ratchet used to test lookback was updated to test expected behavior versus actual behavior.


Bill-Becker:
@lixiangk1 I did the same thing here as I just updated in your PR for REopt.jl at https://github.com/NREL/REopt.jl/pull/128.

Issue #344, opened 2022-08-01
in job/ app, allow modeling multiple outages


adfarth:
* **I'm submitting a ...**
  - [ ] bug report
  - [ ] feature request
  - [ ] support request => Please do not submit support request here, see note at the top of this template.

* **What is the current behavior?**
missing multiple outage modeling inputs and validation for v3 (job/ app)

* **What is the expected behavior?**
add the following ElectricUtilityInputs & required validation: 
  - outage_start_time_steps::Array{Int,1}=Int[],  # we minimize the maximum outage cost over outage start times
  - outage_durations::Array{Int,1}=Int[],  # one-to-one with outage_probabilities, outage_durations can be a random variable
  - outage_probabilities::Array{R,1} where R<:Real = [1.0],
  - outage_time_steps::Union{Nothing, UnitRange} = isempty(outage_durations) ? nothing : 1:maximum(outage_durations),
  - scenarios::Union{Nothing, UnitRange} = isempty(outage_durations) ? nothing : 1:length(outage_durations),

* **What is the motivation / use case for changing the behavior?**
Allow for modeling multiple, probabilistic outages

PR #381, opened 2022-11-03
Multiple outages

hdunham:
- [x] CHANGELOG.md is updated

feature

addresses issue #344 

##### Added
- Added access to the multiple outage stochastic/robust modeling capabilities in REopt.jl. Not all inputs and outputs are exposed, but the following are:
    - **SiteInputs**: **min_resil_time_steps**
    - **ElectricUtilityInputs**: **outage_start_time_steps**, **outage_durations**, **outage_probabilities**
    - **OutageOutputs**: **expected_outage_cost**, **max_outage_cost_per_outage_duration**, **unserved_load_series**, **unserved_load_per_outage**, **microgrid_upgrade_capital_cost**, **generator_fuel_used_per_outage**
- added test using multiple outage modeling
##### Changed
- use TransactionTestCase instead of TestCase (this avoids whole test being wrapped in a transaction which leads to a TransactionManagementError when doing a database query in the middle)

no


zolanaj (approved):
Thank you @hdunham!  I added a couple validation checks in the ElectricUtility `clean()` function in 4e5f2f48b27fd23177b6c3d31ddf5a07ef71ee7e, changed the `OutageOutputs` model key in d10fb38b4127faa5a51043ec1a17f526b4dd402f, and added a few validator tests in 14eceb47d8ce10f7ff31f3ae4eab8544a0170a60.  If you agree with these updates and the merge with develop has tests passing, this should be good to merge.

job/validators.py
```diff
@@ -402,6 +402,16 @@ def update_pv_defaults_offgrid(self):
...
+            if self.models["ElectricUtility"].outage_start_time_steps:
+                max_start_time_step_input = max(self.models["ElectricUtility"].outage_start_time_steps)
+                if max_start_time_step_input > max_ts:
+                    self.add_validation_error("ElectricUtility", "outage_start_time_steps",
+                                                f"Value is greater than the max allowable ({max_ts})")
+
+                if self.models["ElectricUtility"].outage_durations:
+                    if max_start_time_step_input + max(self.models["ElectricUtility"].outage_durations) > max_ts:
+                        self.add_validation_error("ElectricUtility", "outage_durations",
+                                                f"Value is greater than the max allowable ({max_ts} - {max_start_time_step_input})")
```
zolanaj, line 414: One comment that I meant to add in the PR review: since this only includes the "ElectricUtility" model, should this be in that model's `clean()` method in `models.py`?

hdunham: No because it uses max_ts, which is calculated in this function using Settings.time_steps_per_hour

zolanaj: @hdunham thanks for clarifying, I missed that.  This is the right place then, thank you!

PR #382, opened 2022-11-29
Add CHP, /chp_defaults, and /simulated_load end to /dev for v3

Bill-Becker:
See PR https://github.com/NREL/REopt_API/pull/340.


zolanaj (approved):
The changelog update looks good to me, and the lone commit that is in master but not develop is a PR merge with no changed files, so we can merge this (and perhaps do a develop <-- master merge right after).  Thanks @Bill-Becker!

PR #401, opened 2023-01-09
endpoints edits for v3

indu-manogaran:
Edit endpoints to create a working v3 version


Bill-Becker (approved):
@indu-manogaran I added some type-conversion code which replaced some of the stuff you added, but this should handle all load_types (heating and cooling too). 

Also, I removed the /annual_kwh endpoint from reoviews because you can instead use the /simulated_load endpoint to get annual_kwh. That endpoint was still calling the python commercial reference building code anyway, so it would become out-of-sync eventually.

indu-manogaran:
Thanks, Bill. I'll update the UI code to reflect the removal of /annual_kwh.

PR #411, opened 2023-01-31
Use GhpGhx.jl in v2

Bill-Becker:
This PR uses the [GhpGhx.jl](https://github.com/NREL/GhpGhx.jl) Julia package instead of calling Julia scripts/functions within the REopt_API for v2. The previous scripts that were used matched the GhpGhx.jl package, so the functionality is the same. This removes the need for git submodule to differentiate the license that governs the GhpGhx.jl code, as the GhpGhx.jl  package (unregistered) is under a separate license.


Bill-Becker:
@lixiangk1 I just did my final TODO's on this PR, and I switched the name of the branch which is why it looks like a whole new PR even though it's the same as the other one with the `ghp-v3` branch that I just closed. This PR will be necessary for integrating the changes from `GhpGhx.jl` into the v2 API for hybrid GHP. Also FYI, I'm just now adding GHP to the API `job` app (will become "V3") where we are mainly using the REopt.jl Julia package. However, for GHP, we also have the /ghpghx endpoint which exposes the functionality of GhpGhx.jl separately from REopt.jl., so any new inputs or outputs will have to be integrated into /ghpghx models.py as well.

lixiangk1:
Why are the tests failing?

Bill-Becker:
> Why are the tests failing?

Not sure, but I just "re-ran jobs" and looks like it's passing now.

lixiangk1 (approved):

PR #416, opened 2023-02-02
add v3 emissions_profile endpoint

hdunham:
- [x] CHANGELOG.md is updated

extends emissions_profile endpoint feature to v3

emissions_profile url is only available in v1 and v2

dev/emissions_profile url sends back emissions data for given lat/long

no

### Other information:


job/test/test_job_endpoint.py
```diff
@@ -222,6 +222,19 @@ def test_chp_defaults_from_julia(self):
             else:  # Make sure we didn't overwrite user-input
                 self.assertEquals(inputs_chp[key], post["CHP"][key])
 
+    def test_emissions_profile_endpoint(self):
```
Bill-Becker, line 225: Can you move this test to the "test_http_endpoints.py" file which is specifically for testing these endpoints that we are exposing through http.jl? I realize that this is still technically testing a job (/dev) view/endpoint, but I think it's useful to split up tests into categories of 1) running REopt with http.jl `reopt` endpoint and 2) non-reopt endpoints, as we start to add more and more tests.

job/views.py
```diff
@@ -435,3 +435,123 @@ def simulated_load(request):
...
+    except Exception:
+
+        exc_type, exc_value, exc_traceback = sys.exc_info()
+        debug_msg = "exc_type: {}; exc_value: {}; exc_traceback: {}".format(exc_type, exc_value.args[0],
+                                                                            tb.format_tb(exc_traceback))
+        log.error(debug_msg)
+        return JsonResponse({"Error": "Unexpected Error. Please check your input parameters and contact reopt@nrel.gov if problems persist."}, status=500)
+
+
+# def easiur_costs(request):
```
Bill-Becker, line 473: Are we doing this commented function in a future PR?

Bill-Becker (approved):
I'm approving with a suggested re-location of the test you added. Thanks!

Issue #417, opened 2023-02-02
Clean JuMP models post optimization


rathod-b:
**Note: for support questions and enhancement suggestions please use the REopt Forum (under development)**. This repository's issues are reserved for feature requests and bug reports.

* **I'm submitting a ...**
  - [ ] bug report
  - [X] feature request
  - [ ] support request => Please do not submit support request here, see note at the top of this template.

* **Do you want to request a *feature* or report a *bug*?**
Feature request

* **What is the current behavior?**
At present, `GC.gc()` and some model finalization activities are performed post optimization in Julia. However, this can lead to a build-up of memory in a Julia pod/Redis/Celery since Julia isn't the quickest in releasing memory. See https://discourse.julialang.org/t/memory-consumption-growth-with-many-large-milps-in-jump/61895 for further details.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via
https://plnkr.co or similar (you can use this template as a starting point: http://plnkr.co/edit/tpl:AvJOMERrnz94ekVua0u5).

* **What is the expected behavior?**
After each scenario is solved, Julia releases all, or as much memory as it can.

* **What is the motivation / use case for changing the behavior?**
Memory accumulation on production environments can lead to slowdown of REopt API's performance, which can result in degraded performance and loss of time for API users.

* **Please tell us about your environment:**
  
  - REopt API Version: X.X.X
  - Operating System: [all | TypeScript X.X | ES6/7 | ES5 | Dart]
  - Python version: 3.x
  - Julia version: 1.7.3

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)
We could implement `empty!(m)` after each optimization to free up a portion of memory being held by Julia/JuMP. This might not be a solution to the underlying problem but can offer a more time before the production infrastructure gets strained.

This fix was tested in VM using Distributed.jl. Without the fix, the VM ran out of memory after ~2200 PV+BESS reopt scenarios. Memory consumption was at 86% and had reached the read-only-memory in the VM. After the fix, the VM was able to solve ~9000 scenarios in 1 day with memory consumption holding steady at ~70% over 9 Julia worker processes.

PR #418, opened 2023-02-02
Update http.jl

rathod-b:
`empty!()` the JuMP models post solve before `GC.gc()`

- [ ] CHANGELOG.md is updated
- [X] Tests for the changes have been added (for bug fixes / features) (see other information)
- [X] Docs have been added / updated (for bug fixes / features) (see other information)

Memory management feature for JuMP models

Post solve, JuMP models are finalized and handled before garbage collection. However, upon initial testing this does not reduce the size of JuMP models (might even increase the size in some cases).

Now, the JuMP models will be `empty!()` before the `GC.gc()` is called which should free up memory held by these models post solve

No

### Other information:
If all inbuilt API tests pass, testing for this PR will have to be via Rancher. We would have to monitor resource consumption there and see if there is a reduction in how often the infrastructure is out of memory. See Issue #417 for more details on tests done using VM.
No user facing docs need to be updated.

### DIFF
--- a/.helm/templates/julia-deployment.yaml
+++ b/.helm/templates/julia-deployment.yaml
@@ -51,6 +51,4 @@ spec:
             requests:
               cpu: {{ .Values.juliaCpuRequest | quote }}
               memory: {{ .Values.juliaMemoryRequest | quote }}
-            limits:
-              cpu: {{ .Values.juliaCpuLimit | quote }}
-              memory: {{ .Values.juliaMemoryLimit | quote }}
+
--- a/julia_src/http.jl
+++ b/julia_src/http.jl
@@ -20,8 +20,9 @@ function job(req::HTTP.Request)
 	end
     optimizer = backend(m)
 	finalize(optimizer)
-	GC.gc()
     Xpress.postsolve(optimizer.inner)
+	empty!(m)
+	GC.gc()
 	if isempty(error_response)
     	@info "REopt model solved with status $(results["status"])."
     	return HTTP.Response(200, JSON.json(results))
@@ -115,8 +116,11 @@ function reopt(req::HTTP.Request)
 	if typeof(ms) <: AbstractArray
 		finalize(backend(ms[1]))
 		finalize(backend(ms[2]))
+		empty!(ms[1])
+		empty!(ms[2])
 	else
 		finalize(backend(ms))
+		empty!(ms)
 	end
     GC.gc()
 


Bill-Becker (approved):

rathod-b:
Commented out some Julia pod memory management safeguards implemented by Nick M. for testing purposes.

Bill-Becker:
@rathod-b I'm curious to see if this helps our current issues with calling /dev on production right now. It's super slow, and it looks like /dev (REopt.jl with 2-threads) is causing a quicker buildup of memory on the pods/containers. I don't see how this could hurt, so I'd like to merge this in soon (by mid-day Monday to test).

I noticed you removed the memory limit setting for the Julia pods - do we think that is necessary and safe, or did you do this just to test?

Bill-Becker:
Hosting the API locally, I did a pretty manual tracking of memory in the julia container as I ran several runs, before and after these additions with empty! - I'm still seeing the same kind of memory growth. Whenever http.jl (and therefore Julia) is restarted, the memory consumption reduces back down to small, but then I do runs and it creeps back up. Was the improvement you saw pretty conclusive? Let's chat early next week.

rathod-b:
> Hosting the API locally, I did a pretty manual tracking of memory in the julia container as I ran several runs, before and after these additions with empty! - I'm still seeing the same kind of memory growth. Whenever http.jl (and therefore Julia) is restarted, the memory consumption reduces back down to small, but then I do runs and it creeps back up. Was the improvement you saw pretty conclusive? Let's chat early next week.

I believe the improvement I saw after adding empty!() was quite conclusive because prior to that change, REopt.jl would have maxed out the VM memory in ~2k runs. After the change it didnt max out the memory. It got relatively close to using 70% of RAM on the VM and stayed near that number for the runtime of about 2 days (with interruptions in the VM).

Maybe we should explore what it would look like to host REopt.jl on a dedicated HPC node? everyone is moving on to the new Kestrel system so there is reduced load on Eagle now.

Bill-Becker:
Ok, yeah I was seeing the memory would grow from less than 1GB to up to 4.5-5GB in less than 10 runs, and then it did seem to stop growing as much. Sometimes it would reduce by a little, then grow a little more. I didn't do nearly as exhaustive testing as you did, but there was a clear trend of growth that wouldn't reduce. Do you understand why it grows initially and then seems to level off, as it did in your case and mine?

rathod-b:
> Ok, yeah I was seeing the memory would grow from less than 1GB to up to 4.5-5GB in less than 10 runs, and then it did seem to stop growing as much. Sometimes it would reduce by a little, then grow a little more. I didn't do nearly as exhaustive testing as you did, but there was a clear trend of growth that wouldn't reduce. Do you understand why it grows initially and then seems to level off, as it did in your case and mine?

Unsure as to why memory would grow initially. It could be Julia and Xpress expanding memory consumption for problem solving, and then eventually they find a steady state condition where memory becomes a limiting factor for all the workers?

Bill-Becker:
@rathod-b any reason not to merge this into develop first instead of master? And I think this is good to go, right?

Bill-Becker:
This was merged into develop already, so maybe we can close this?

rathod-b:
> This was merged into develop already, so maybe we can close this?

Yes, we sure can! Apologies for the delayed responses on this.

PR #420, opened 2023-02-06
Add db index for MessageModel and PVModel

Bill-Becker:
This is the SQL translated from django's migration file for the two db tables:
```
BEGIN;
--
-- Alter index_together for messagemodel (1 constraint(s))
--
CREATE INDEX "reo_messagemodel_run_uuid_f98c3013_idx" ON "reo_messagemodel" ("run_uuid");
--
-- Alter index_together for pvmodel (1 constraint(s))
--
CREATE INDEX "reo_pvmodel_run_uuid_ffc792f5_idx" ON "reo_pvmodel" ("run_uuid");
COMMIT;
```


GUI (approved):
I'm obviously not super familiar with the Django side of this, but that SQL looks right, so this looks good to me, and I think these indices will help with some of the production load. Thanks!

PR #425, opened 2023-02-10
Add GHP to /dev for v3

Bill-Becker:
- [x] CHANGELOG.md is updated

Add GHP to /dev for v3

No GHP in /dev for v3

Yes GHP in /dev for v3

No

### Other information:
Updated julia_src Julia environment to use v1.8, as well as in the Dockerfile


job/src/process_results.py
```diff
@@ -33,7 +33,7 @@
                         ElectricUtilityInputs, ExistingBoilerOutputs, CHPOutputs, CHPInputs, \
                         ExistingChillerOutputs, CoolingLoadOutputs, HeatingLoadOutputs,\
                         HotThermalStorageOutputs, ColdThermalStorageOutputs, OutageOutputs,\
-                        REoptjlMessageOutputs
+                        REoptjlMessageOutputs, GHPInputs, GHPOutputs
```
zolanaj, line 36: `GHPInputs` isn't used, but I think it should be.  We want it to update the inputs to `space_heating_efficiency_thermal_factor` and `cooling_efficiency_thermal_factor` after defaults are provided in Julia, like for CHP in `update_inputs_in_database` below - correct?

job/models.py
```diff
@@ -5623,6 +5623,280 @@ class CoolingLoadOutputs(BaseModel, models.Model):
...
+        ],
+        blank=True,
+        help_text="Maximum utility rebate"
+    )
+
+
+    def clean(self):
+        # perform custom validation here.
+        pass
+
```
zolanaj, line 5880: Should we add something here that checks for either `ghpghx_responses` or something else in the inputs dictionary?  Or would that be caught upstream in `ghpghx/models.py`?

zolanaj (changes requested):
Thanks @Bill-Becker, endpoints look good!  There are a couple questions here related to validation and updating results.  This former might be ok as is, but I believe we probably need the latter in place to be consistent with other techs.

job/models.py
```diff
@@ -5623,6 +5623,280 @@ class CoolingLoadOutputs(BaseModel, models.Model):
...
+    )
+
+    ghpghx_responses = ArrayField(
+        models.JSONField(
+            null=True,
+            editable=True
+        ),
+        default=list, blank=True, null=True,
+        help_text=("ghpghx app response(s) to re-use a previous GhpGhx.jl run without relying on a database entry")
+    )
```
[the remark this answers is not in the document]
Bill-Becker: Good call, I added handling of ghpghx_response_uuids in 0ea7fb4d8ba1c27d6af611c641fabae0c17bdc83. I tested this locally by creating a ghp_uuid entry in the /ghpghx database, and then I POSTed a run to /dev with GHP.ghpghx_response_uuids, and it worked. I had to delete the ghpghx_response_uuids from the POST because it was an unexpected argument for the GHP object in REopt.jl, and while ideally it would still be recorded in the inputs, I think this is fine for now. In theory I would add a unit test for this, but that would get a little tricky because the test suite uses a temporary database so the /ghpghx uuid would have to be created first and somehow stored before the test could be run. This is doable, but since this is a pretty rare use case, it's not worth the effort right now.

job/src/process_results.py
Bill-Becker: Agreed, I did this in cab9a20ab1669a93c87d2ee08ea9a9d794c52c7c.

job/models.py
Bill-Becker: I think you're referring validation of ghpghx_responses input, but if the user is using ghpghx_responses, we assume that this is a direct/unmodified response from/ghpghx so it should be good. The inputs validation for ghpghx_inputs that is used to create the ghpghx_responses happens when either the user or REopt calls the /ghpghx endpoint. If the user modifies ghpghx_responses, we don't really do validation of that before passing to REopt like we do for urdb_responses (I think we do that). Let me know if I'm misinterpreting your request/question.

job/models.py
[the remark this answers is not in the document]
zolanaj: This is exactly the test case and addition to `clean()` I was thinking of - thank you!  I think that given the specific use cases intended for GHP I agree this is sufficient, and we can revisit if an issue comes up from users trying to do this for some reason.

job/src/process_results.py
zolanaj: Looks good, thank you!

job/models.py
zolanaj: Yes, this and what you did in https://github.com/NREL/REopt_API/commit/0ea7fb4d8ba1c27d6af611c641fabae0c17bdc83 makes sense - thanks for clarifying.

zolanaj (approved):
Thanks for addressing the comments @Bill-Becker!  This is good as-is and I'll wait until you're back before merging.  I'll update the changelog and wait on the rest.

PR #429, opened 2023-02-26
Return has_stacktrace true/false value

rathod-b:
- [X] CHANGELOG.md is updated

Feature

Currently, when REopt API returns an error, there is no metadata along the error indicating if it has a stacktrace or not.

Now, REopt API will return a `has_stacktrace` field which will be true if the error had a stacktrace. otherwise it will be `false`.

No, this field is in addition to the existing REopt API job app models, and should not break any existing code.

### Other information:
The has_stacktrace field is a Boolean Field which defaults to False. If this field is true, the user can access the stack trace returned from REopt.jl via the API as denoted in these documents: https://github.com/NREL/REopt-API-Analysis/wiki/Common-Errors.

`results["messages"]["errors"]` will contain a dictionary of error (i.e. key) and error message or stacktrace (i.e. value) 

This PR depends on https://github.com/NREL/REopt.jl/pull/190 getting merged into REopt.jl


rathod-b:
This PR depends on https://github.com/NREL/REopt.jl/pull/190 getting merged into REopt.jl master branch.

job/models.py
```diff
@@ -4548,6 +4548,12 @@ class REoptjlMessageOutputs(BaseModel, models.Model):
         default = list,
     )
 
+    has_stacktrace = models.BooleanField(
+        blank=True,
+        default=False,
+        help_text="True if the error message has a stacktrace, false otherwise"
```
adfarth, line 4554: @rathod-b I think it could be helpful if you add a brief note as to why some error messages might have slack traces and some not

adfarth (approved):
@rathod-b Looks good to me! I think the PR description should be modified to no longer say this is a text field, right? Also, can a user access the slack trace if has_slacktrace is True? If so, could we include info in this PR description as to how to do that, or why this was necessary?

rathod-b:
> @rathod-b Looks good to me! I think the PR description should be modified to no longer say this is a text field, right? Also, can a user access the slack trace if has_slacktrace is True? If so, could we include info in this PR description as to how to do that, or why this was necessary?

Yes a user can access the stacktrace if it is returned from REopt.jl. I have added an example of how to do that in the PR description. I have also updated documentation on error messages under the last section here (https://github.com/NREL/REopt-API-Analysis/wiki/Common-Errors)

job/models.py
rathod-b: Updated! Please take a look

PR #430, opened 2023-02-27
Add boiler and steamturbine to API

rathod-b:
- [x] CHANGELOG.md is updated
- [x] Docs have been added / updated (for bug fixes / features)*
 * documentation added in `reoptjl/models.py`. Apart from that, these new features are documented in webtool manual.

Feature:
- Boiler inputs and outputs
- Steamturbine inputs and outputs
- Steamturbine defaults endpoint

Currently, REopt.jl's Boiler and SteamTurbine capabilities are not accessible through REopt API. This PR makes them accessible.

Not a feature change, but new features

No


rathod-b:
Question for the reviewer(s), does upgrading http.jl from v0.9.17 break the `@register` macro used in http.jl? I was getting that error and had to revert the upgrade of http.jl

rathod-b:
We should also take stock of where `MAX_BIG_NUMBER` is used and if we need multiple constants to represent various maximum values. (i.e using 1.0e8 in steam turbine vs 1.0e9 in generators)

Bill-Becker:
> Question for the reviewer(s), does upgrading http.jl from v0.9.17 break the `@register` macro used in http.jl? I was getting that error and had to revert the upgrade of http.jl

Yes it does! We actually updated HTTP.jl package in the erp-main branch and then had to change @register to register!.

Bill-Becker:
@rathod-b I combined /steamturbine_defaults into /chp_defaults because that's the way v2 works, it's mainly used for the webtool, and I think we'll create a lot more work for the webtool developers if we change it. My last commit was an attempt to finish the test_steamturbine_defaults_from_julia test, but currently the REopt.jl run is erroring. Can you pick up from here? I think we're really close now to wrapping this up.

rathod-b:
It seems like there is a misalignment in *MAX_BIG_NUMBER* in `models.py`. The Julia package is using 1.0e9 as the upper limit for `max_mmbtu_per_hour` whereas the API is using MAX_BIG_NUMBER (1.0e8). Should we make multiple max big numbers? Is there a separate issue to address this?

rathod-b:
Note to change base to merge into from master to develop.

Bill-Becker:
> We should also take stock of where `MAX_BIG_NUMBER` is used and if we need multiple constants to represent various maximum values. (i.e using 1.0e8 in steam turbine vs 1.0e9 in generators)

I agree - I don't know of any reason we use e8 for something while e9 for other big numbers.

julia_src/Manifest.toml
Bill-Becker, line 1: Looks like we're going backward slightly on package versions for the most part. Any thoughts on using the develop Manifest.toml, or would merging develop into revive_Boiler use the develop Manifest.toml?

Bill-Becker (approved):
I don't have an issue with the slight regression of package versions, but maybe consider my comment on that.

PR #431, opened 2023-03-01
Add summary URL to API v3

rathod-b:
- [X] CHANGELOG.md is updated
- [ ] Tests for the changes have been added (for bug fixes / features) (see other information)
- [ ] Docs have been added / updated (for bug fixes / features) (see other information)

Feature

At present, there is no capability in API v3 to:
1. Save user_uuid and webtool_uuid
2. Query user REopt summary for all runs or by chunk.
3. Unlink a run_uuid from a user_uuid

Upon completion of this PR, a provided `user_uuid` can be used to pull all scenarios and return summary metadata using a `summary` or `summary_by_chunk` endpoint in API v3.
A webtool run can also be unlinked from a user_uuid in API v3

No

### Other information:
- Testing for this capability will be done directly with user interface, hence no programmatic tests added
- No user facing documents to update for this back-end endpoint


rathod-b:
It would be great to get a review of the approach I am taking in teleporting the `summary` for API v3 in views.py. The current code produces the following output in Django for the fields that have thus far been coded:
image

rathod-b:
Created a working endpoint which returns the following fields:
image

Need to still
- Test this with user_uuid and multiple associated runs, 
- Handle unlinked scenarios
- Handle fields which might not be present in the model, should they be 0s?

rathod-b:
Tested and validated all three endpoints locally by adding a user_uuid. However I could use some help in writing up a testset for these endpoints.

job/test/test_job_endpoint.py
```diff
@@ -311,3 +311,11 @@ def test_peak_load_outage_times(self):
         self.assertHttpOK(resp)
         outage_start_time_steps = json.loads(resp.content)["outage_start_time_steps"]
         self.assertEquals(outage_start_time_steps, expected_result)
+    
+    def test_summary_endpoints(self):
+
+        user_uuid = "501d1dd9-9779-470b-a631-01c5fbdee570"
+        resp = self.api_client.get(f'/dev/user/{user_uuid}/summary')
```
Bill-Becker, line 318: @rathod-b This is where I thought we needed to use python requests instead of the test api_client because my understanding is that every test uses an temporary database that gets deleted after each test.

Bill-Becker: Honestly, I'm not super worried about creating a test right now. I think the real "test" will be making sure this can be used with the UI which is our main goal right now. I attempted to hook this API branch up to the UI and started a fix a couple of issues on the UI side before I decided to give up and hand it over to Erika to do that. Let's wrap up any known issues and get this branch deployed to the development server so she can start testing it out.

Bill-Becker:
If you are good with the state of this branch after Erika's testing of it, I'm good to merge into develop and then we'll merge this with the branch we are using for the ERP UI beta (and merge into master whenever the next master <- develop merge happens, but that's less pressing because the production UI doesn't use this).

Bill-Becker (approved):
Feel free to merge whenever you are comfortable with the state between UI and the this branch.

PR #433, opened 2023-03-06
change ubuntu-18.04 to ubuntu-latest in github actions

hdunham:
- [x] CHANGELOG.md is updated
- [ ] N/A Tests for the changes have been added (for bug fixes / features) 

Updated ubuntu-18.04 to ubuntu-latest in GitHub push/pull tests because 18.04 was deprecated in GitHub Actions

all GitHub actions are automatically cancelled during "awareness brownouts" and will soon always fail

github actions run again

no


zolanaj (approved):
My only note here is that if we run into a similar situation in the future, a possible fix will be to revert an older (but not deprecated) version of ubuntu if a breaking change for our test suite is introduced.  I think 'latest' is still the best path forward though.  Thank you @hdunham for the fix!

PR #436, opened 2023-03-10
Add /dev/schedule_stats endpoint

Bill-Becker:
Reuse the same `reo` view `schedule_stats` function for `/dev/schedule_stats` because this endpoint is purely a user-interface helper endpoint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,35 +26,26 @@ Classify the change according to the following categories:
     ##### Removed
     ### Patches
 
-## Develop - 2023-03-06
+## Develop - 2023-03-10
 ### Minor Updates
-### Added
+#### Added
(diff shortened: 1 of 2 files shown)


Bill-Becker:
@indu-manogaran  note, this does not include the CHP size_class starting at 0 (again), which currently lives in a branch of REopt.jl. I thought it would be cleaner to just merge this first, and the CHP size_class update requires no API changes, so that would just get updated the next time we update REopt.jl in the API. Separately, I have a goal to merge all the erp API branch into master and deploy next tuesday with Hallie. That will have the CHP size_class update in it as well.

indu-manogaran (approved):

indu-manogaran:
@Bill-Becker Got it. I've verified the endpoint addition.

PR #451, opened 2023-03-29
Ghp hybrid

lixiangk1:
- [ ] CHANGELOG.md is updated

Adds hybrid functionality to REopt's GHP model


Manifest.toml
Bill-Becker, line 1: Let's remove this. Not sure where it came from.

Project.toml
Bill-Becker, line 1: Remove

reo/tests/test_ghp_hybrid.py
```diff
@@ -0,0 +1,79 @@
...
+
+        ghp_uuid = d["outputs"]["Scenario"]["Site"]["GHP"]["ghp_chosen_uuid"]
+        print("GHP uuid chosen = ", ghp_uuid)
+
+        # Output number of boreholes and heatpump sizing 
+        n_boreholes = d["outputs"]["Scenario"]["Site"]["GHP"]["ghpghx_chosen_outputs"]["number_of_boreholes"]
+        heatpump_tons = d["outputs"]["Scenario"]["Site"]["GHP"]["ghpghx_chosen_outputs"]["peak_combined_heatpump_thermal_ton"]
+        # Comparison to TESS exe range
+        # TODO: Number of boreholes vary between runs with the same input
+        # self.assertAlmostEqual(n_boreholes, 45)
```
Bill-Becker, line 78: Seems like a test for hybrid should test something about either the GHX size, or if that is not reliable, it should test that there is some auxiliary heating or cooling, depending on the input scenario. The heat pump size is the same for hybrid or not, right?

Bill-Becker (commented):
A couple of files to delete, and I'd suggest we make the hybrid GHP test check for something hybrid-specific, like auxiliary heater/cooler energy/cost instead of just the heat pump size which doesn't change with hybrid.

reo/tests/test_ghp_hybrid.py
lixiangk1: @Bill-Becker I pushed a change to delete the toml files but it seemed to have updated some other stuff? Do I need to undo something?

I'm also not sure about how to do the hybrid GHP tests... since boreholes changes with different iterations, it seems that a lot of the aux calculations change slightly too.... I basically removed checks that were changing until only heat pump size was left, which is not a hybrid var...

Bill-Becker (commented):
I merged develop and updated the Julia environment (GhpGhx and REopt packages, but it also updated some other dependencies), so that updated the Manifest.toml - is that what you meant by "it seemed to update some other stuff"?

reo/tests/test_ghp_hybrid.py
Bill-Becker: Looks good!

Bill-Becker (approved):

PR #474, opened 2023-07-11
Change app name "job" to "reoptjl"

Bill-Becker:
The current name "job" for the app which will be the API V3 and mainly calling the REopt.jl Julia package is ambiguous and confusing. This PR renames the "job" app to "reoptjl" to be more intuitively named. 

I followed this procedure to change the name of a Django app, using a package called django_rename_app (but a forked version of that package that is compatible with Python v3.8) that has to be installed temporarily until the merged name change is deployed once on the production server (after that, the package can be deleted):

https://odwyer.software/blog/how-to-rename-an-existing-django-application

I tested the merge of the name change both locally and on the development server, where I merged this branch into a copy of the develop branch with an existing database that had the old name (job). We will obviously need to closely monitor the deployment when we merge into production.


Bill-Becker:
@adfarth Here are the steps to test that the change works by merging it into an existing database:

1. Clear your docker containers and database volume: docker compose down -v
2. Switch to the develop branch
3. Spin up the develop branch: docker compose build and then docker compose up
4. Run REopt once on your local API (POST to dev/job endpoint) to create an entry in your local database
5. Stop and remove the containers, but don't delete the database volume: docker compose down
6. Switch to this branch
7. Spin up the API: docker compose build (required because there is a new package installed in the Dockerfile), then docker compose up; it should make all the migrations, and in the background it should run that django rename package.
8. Run REopt once on your local API

Let me know if you run into issues!

adfarth:
> @adfarth I figured why the steps below were not replicating how the API gets deployed and therefore why it was failing locally but succeeding on the development server. When you redeploy updates to a branch that already exists and is running on the server (and that already has a database), the django service stays active and it has already run the db migrations. Doing docker compose down removes the django service, and just keeping the volume doesn't prevent django from **needing to run the migrations again**. I had put the `rename app` command before running the migrations in the docker-compose.yml file, and the `rename app` command expects the migrations to already have been run. That's why it was erroring that things were missing. Assuming your API is in the same format as it was before, do this:
> 
> 1. Clear your docker containers and database volume: docker compose down -v
> 2. Switch to the develop branch, and docker compose up (note, if you last built the reoptjl-app branch, you will still have the extra package installed from the last time you built that, and no need to do `build` again when you switch back to reoptjl-app branch).
> 3. **Stop the containers (cntrl/cmd+c), but do not delete them (don't do docker compose down)**
> 4. Switch to this branch
> 5. Run docker compose up; **you should see a bunch of "Renamed..." in the django container logs - that's it running the `rename app` package command
> 6. Run REopt once on your local API
> 
> Let me know if you run into issues!

These steps worked with the exception that I ended up having to docker compose build before docker compose up in Step 5

CHANGELOG.md
```diff
@@ -45,16 +45,16 @@ Classify the change according to the following categories:
 ## v2.12.0
 ### Minor Updates
 ##### Added
```
adfarth, line 47: @Bill-Becker  is this intentional to replace "job\" with "reopjl/" throughout the Changeling?

Dockerfile
```diff
@@ -15,6 +15,7 @@ COPY . /opt/reopt
 # Install python packages
 WORKDIR /opt/reopt
 RUN ["pip", "install", "-r", "requirements.txt"]
+RUN ["pip", "install", "django-rename-app@git+https://github.com/tonioo/django-rename-app@10c87d62bfc14853379f6a0505f53ba4532848ab"]
```
adfarth, line 18: @Bill-Becker will this line need to be reverted once merged into master?

adfarth (approved):
@Bill-Becker I am approving because the Docker process was completed successfully. I left two comments in the code and am wondering if in addition to deleting the django_rename_app package, there are any other changes that will need to be reverted once merged in to master and deployed on the production server?

Dockerfile
Bill-Becker: Yes, good point

CHANGELOG.md
Bill-Becker: Good question, because I could see it going either way, but I feel like if we are renaming the app then we'd want to change it in the changelog because people reading the changelog would know what app we're referring to. Did you see someplace where it didn't look right? The only weird place where `job` is still relevant is the `job/<run_uuid>/results` URL which is where you grab the results, so I think as long as I didn't replace that we should be good. Also, the `test_job_endpoint.jl` is still properly named because it's referring to the `reoptjl` app's `/job/results` endpoint

PR #486, opened 2023-08-10
Update ExistingChiller cop inputs from Julia

Bill-Becker:
Hot fix for adding `ExistingChiller.cop` to API response inputs when relying on REopt.jl logic for the default value (dependent on cooling load).


zolanaj (approved):
Thank you @Bill-Becker!

PR #496, opened 2023-09-06
Update chiller cop call

zolanaj:
- [x] CHANGELOG.md is updated

Bug fix

The `get_existing_chiller_default_cop` endpoint only accepts posts with every key populated

- The `get_existing_chiller_default_cop` endpoint now had all optional inputs; default values are already in place in REopt.jl in the event that `None`/`nothing` is passed to Julia.  **edit 9/7: a default is now in place for the max thermal factor.**
- A new test shows that when no inputs are provided the "unknown" chiller default cop is returned.
- The thermal max load input is in tons, rather than kW; the unit conversion is then performed in the API before calling the REopt.jl function (which converts back to tons).  

No breaking changes

### Other information:
This should allow for only `max_load_kw` or `max_load_kw_thermal` but not both to be populated in UI calls (which are necessary for the populating chiller COPs for "Upload" and "Custom" cooling loads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ Classify the change according to the following categories:
     ##### Removed
     ### Patches
 
+## Develop 2023/09/06
+### Minor Updates
+##### Fixed
+- Fixed a bug in the `get_existing_chiller_default_cop` endpoint not accepting blank/null inputs that are optional
+
 ## v2.15.0
 ### Minor Updates
 ##### Added
--- a/reoptjl/test/test_http_endpoints.py
+++ b/reoptjl/test/test_http_endpoints.py
@@ -218,9 +218,21 @@ class TestHTTPEndpoints(ResourceTestCaseMixin, TestCase):
             "max_load_kw_thermal":100
         }
 
-        # Call to the django view endpoint /ghp_efficiency_thermal_factors which calls the http.jl endpoint
+        # Call to the django view endpoint /get_existing_chiller_default_cop which calls the http.jl endpoint
         resp = self.api_client.get(f'/dev/get_existing_chiller_default_cop', data=inputs_dict)
         view_response = json.loads(resp.content)
         print(view_response)
 
         self.assertEqual(view_response["existing_chiller_cop"], 4.4)
+
+        # Test empty dictionary, which should return unknown value
+        inputs_dict = {}
+
+        # Call to the django view endpoint /get_existing_chiller_default_cop which calls the http.jl endpoint
+        resp = self.api_client.get(f'/dev/get_existing_chiller_default_cop', data=inputs_dict)
+        view_response = json.loads(resp.content)
+        print(view_response)
+
+        self.assertEqual(view_response["existing_chiller_cop"], 4.545)
+
+        
--- a/reoptjl/views.py
+++ b/reoptjl/views.py
@@ -596,6 +596,7 @@ def get_existing_chiller_default_cop(request):
         inputs_dict = {}
         for key in ['existing_chiller_max_thermal_factor_on_peak_load','max_load_kw','max_load_kw_thermal']:
             inputs_dict[key] = request.GET.get(key)
+            # allow empty keys to be None, convert all other inputs to floats
             if inputs_dict[key] is not None:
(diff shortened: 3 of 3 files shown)


Bill-Becker (changes requested):
Looks like this always provides the "unknown" cop value if `existing_chiller_max_thermal_factor_on_peak_load` is not provided. I'm thinking that should be optional because we supply the default 1.25 in REopt.jl, so we should have that as the default. Also, I think the units for the thermal input should be ton instead of kW because if they have the thermal cooling load data, that would likely be in units of ton.

Bill-Becker:
Looks like I was the one who made "if existing_chiller_max_thermal_factor_on_peak_load == nothing then return unknown cop" in REopt.jl - so as the shortcut solution, maybe just assign the default existing_chiller_max_thermal_factor_on_peak_load of 1.25 within the API view, if not supplied, instead of going back and changing REopt.jl?

Bill-Becker:
Sorry for the string of messages. I was looking at the wrong version of REopt.jl and did not see the updates you recently made to `get_existing_chiller_default_cop()` function. Ignore my last message because it looks like the function should return one of the two "known" COPs if the user only inputs max_load_kw OR max_load_kw_thermal, without existing_chiller_max_thermal_factor_on_peak_load. But, for some reason I'm not seeing that in the API view behavior.

Bill-Becker:
I see why it returns the unknown cop with no max_thermal_factor (the OR statement):

https://github.com/NREL/REopt.jl/blob/3f2109ae30ea4bcb725c21d7aea7595f9d343d5b/src/core/heating_cooling_loads.jl#L230

In REopt.jl, the max_thermal_factor always gets assigned from the existing_chiller (either from default or user input) in scenario.jl. I'd suggest we just add this default in the API endpoint as the quick workaround, so the endpoint user (webtool) does not have to put this in.

zolanaj:
> I see why it returns the unknown cop with no max_thermal_factor (the OR statement):
> 
> https://github.com/NREL/REopt.jl/blob/3f2109ae30ea4bcb725c21d7aea7595f9d343d5b/src/core/heating_cooling_loads.jl#L230
> 
> In REopt.jl, the max_thermal_factor always gets assigned from the existing_chiller (either from default or user input) in scenario.jl. I'd suggest we just add this default in the API endpoint as the quick workaround, so the endpoint user (webtool) does not have to put this in.
> 
> Let's also change the max_load_kw_thermal input for the API endpoint to be max_load_ton and do the unit conversion in the endpoint.

I refactored how inputs are handled since each key is handled in a different way (one is just a float conversion if not None, one has a type conversion and a unit conversion, and one has a default).  I'll request a new review once tests pass.

In light of the underlying cause you discovered, should we maybe make max_thermal_factor a required input in REopt.jl instead of allowing nothing since it would never be covered?  (We could still use the unknown COP if both max_load_kw and max_load_kw_thermal are nothing, but I am not sure why you'd call the endpoint without at least one of those values known.)

Bill-Becker:
I don't think we want to make max_thermal_factor a required input - I like the 1.25 default. What do you mean by " it would never be covered"?

zolanaj:
> I don't think we want to make max_thermal_factor a required input - I like the 1.25 default. What do you mean by " it would never be covered"?

I'm just referring to the case of "nothing" being passed into max_thermal_factor - sorry if that was unclear.  (in other words, let 1.25 be the default instead of nothing, and don't check for isnothing(max_thermal_factor) when determining whether or not to use the unknown COP.

zolanaj:
> > > I don't think we want to make max_thermal_factor a required input - I like the 1.25 default. What do you mean by " it would never be covered"?
> > 
> > 
> > I'm just referring to the case of "nothing" being passed into max_thermal_factor - sorry if that was unclear. (in other words, let 1.25 be the default instead of nothing, and don't check for isnothing(max_thermal_factor) when determining whether or not to use the unknown COP.
> 
> That makes sense except we don't want the default to live in two places, the existing_chiller and the get_existing_chiller_cop() function.

Agreed.  I think what we have here works fine in that case; I don't see a clear way of referencing the default from Scenario.jl when calling the endpoint form the API and I'm not sure we need to do so.

PR #499, opened 2023-09-13
Add v3 URLs

Bill-Becker:
Add the /v3 URLs which copies over all the /dev URLs from the reoptjl app (futurecosts app will stay in /dev).

This has been tested against the erp-v3 webtool locally with /dev changed to /v3 for all API calls.


Bill-Becker:
Just need to do this once we want to make /v3 the /stable API:

image
and
image

PR #502, opened 2023-09-18
Update chiller cop call

zolanaj:
- [ ] CHANGELOG.md is updated

Allows for an empty string to be passed as a value at the 

empty or "nil" values are passed as an empty string which yields a type conversion error at the `get_existing_chiller_default_cop` endpoint

Checks for `None` have been extended to include checks for empty strings as well

No breaking changes


zolanaj:
@Bill-Becker this is a quick PR to allow for easier message passing from the UI, as Erika ran into errors trying to pass null values ("nil" in the UI) through the default electric chiller COP endpoint.

Bill-Becker (approved):

PR #506, opened 2023-09-20
Add Outages results, Make Outages load balance == instead of >=

Bill-Becker:
- Add the critical load in the Outages results for all modeled outages
- Add Wind results for outages, just like PV - it was using Wind in the load balance, but not reporting in the results
- Add battery soc series results for all modeled outages
- Make the outage load balance `supply == load` instead of `supply >= load`, so runs with outages **may** have slightly different results (very slightly) from the change to the model


Bill-Becker:
@hdunham I believe the change from `>=` to `==` for the outages load balance could very well change the result by the amount that the `test_erp.test_erp_with_reopt_run_uuid` test is failing by (very small).

PR #513, opened 2023-09-28
Defaults updates fy23

adfarth:
### Please check if the PR fulfills these requirements
- [x] CHANGELOG.md is updated
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features) --> User manual will be updated. Markdown text updated in this PR

### What kind of change does this PR introduce?
(Bug fix, feature, docs update, ...)
Annual update to default values in REopt. Updated to reflect the latest available data (largely, cost data). 

### Does this PR introduce a breaking change?
(What changes might users need to make in their application due to this PR?)
Users will get different results due to updates to defaults. Users can override default values with old values to obtain the same results. See table below. 

### Defaults updates in this PR:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/afarthin/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/afarthin/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">

<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">

Input   name (manual) | Reopt Input Key | Reopt input   name | Current Default | New Default
-- | -- | -- | -- | --
Owner Discount   rate, nominal (%) | Financial | owner_discount_rate_fraction | 0.0564 | 0.0638
Offtaker   Discount rate, nominal (%) | Financial | offtaker_discount_rate_fraction | 0.0564 | 0.0638
Electricity cost escalation rate, nominal (%) | Financial | elec_cost_escalation_rate_fraction | 0.019 | 0.017
Existing boiler fuel cost escalation rate, nominal (%) | Financial | existing_boiler_fuel_cost_escalation_rate_fraction | 0.034 | 0.015
Boiler fuel cost escalation rate, nominal (%) | Financial | boiler_fuel_cost_escalation_rate_fraction | 0.034 | 0.015
[...]

PR #509, opened 2023-09-25
enable hybrid ghx

rathod-b:
### Please check if the PR fulfills these requirements
- [X] CHANGELOG.md is updated
- [?] Tests for the changes have been added (for bug fixes / features)
- [?] Docs have been added / updated (for bug fixes / features)

### What kind of change does this PR introduce?
(Bug fix, feature, docs update, ...)
Features (2)

### What is the current behavior?
(You can also link to an open issue here)
At present, user is unable to access hybrid GHX or centralized GHP capabilities built in REopt.

### What is the new behavior (if this is a feature change)?
User is able to access hybrid GHX and centralized GHP capabilities built in REopt.jl

### Does this PR introduce a breaking change?
(What changes might users need to make in their application due to this PR?)
If user is missing some inputs required by GHP tech (like **heat_pump_configuration**) their GHPGHX responses might fail.

PR #514, opened 2023-09-29
v3.1.0 Defaults Updates and Hybrid+Central GHP

adfarth:
- Annual default values update. See https://github.com/NREL/REopt_API/pull/513
- Add Hybrid and Central GHP in https://github.com/NREL/REopt_API/pull/509
- Fix proforma metrics for GHP and add thermal results in https://github.com/NREL/REopt_API/pull/521


Bill-Becker (approved):
This has both the FY23 defaults updates which we'll need to deploy at the same time as the web tool updates defaults, as well as the Hybrid and Central GHP which we need to merge Monday to still hit the FY23 milestone.

Bill-Becker:
@lixiangk1 I missed this in the GHP PR review, but looks like we need to add Central GHP fields for ghpghx models.py GHPGHXInputs model. We should add a test for central GHP which would have caught this.

Bill-Becker (changes requested):
@lixiangk1 I think there needs to be Central GHP fields added to the GHPGHXInputs model in the ghpghx app and ideally add a test to confirm all GHPGHXInputs and GHPGHXOutputs are synced up with GhpGhx.jl - especially for central. I missed this in the PR into develop for GHP hybrid and central.

Bill-Becker (approved):
@lixiangk1 made the updates for GHP, so approving this (again). We just need to sync with the web tool deploy for the defaults updates.

PR #516, opened 2023-10-02
Debug central ghp

lixiangk1:
Add central plant GhpGhx.jl variables to the API model; add central plant test


reoptjl/test/test_job_endpoint.py
```diff
@@ -283,4 +283,20 @@ def test_hybridghp(self):
...
+        resp = self.api_client.post('/v3/job/', format='json', data=post)
+        self.assertHttpCreated(resp)
+        r = json.loads(resp.content)
+        run_uuid = r.get('run_uuid')
+
+        resp = self.api_client.get(f'/v3/job/{run_uuid}/results')
+        r = json.loads(resp.content)
+
+        # calculated_ghx_residual_value 117065.83
+        self.assertAlmostEqual(r["outputs"]["Financial"]["lifecycle_capital_costs"], 1046066.8, places=0)
```
Bill-Becker, line 302: Do we expect this to pass all the time with the variability in GHX sizing? Also, since we just have a temporary fix which will change costs with rebate $/ton incentives, should we force in that there are no rebate incentives in the inputs in this tests post (just to be certain things don't change later)?

lixiangk1: @Bill-Becker I updated the hybrid and central GHP tests with a bigger tolerance. Tests seem to be passing. Do you see any other issues?

Bill-Becker (approved):

PR #553, opened 2023-12-21
Add proforma to API v3

rathod-b:
- [ ] CHANGELOG.md is updated


proforma_v3_test/create-proforma.ipynb
```diff
@@ -0,0 +1,1803 @@
...
+    "     if row['Type'] != '0':\n",
+    "        inandout_sheet.write(\n",
+    "            'A'+str(index+1),\n",
+    "            row['Field'],\n",
+    "            d[row['Type']]\n",
+    "        )\n",
+    "\n",
+    "inandout_sheet.write('B1', '', d['header_format_right'])\n",
+    "\n",
+    "inandout_sheet.write('B2', '', d['subheader_format_right'])\n",
```
adfarth, line 436: Could probably for-loop this section?

adfarth:
Summary of some initial testing: 
- 3/3 testing results jsons (attached below) produced errors in the code 
[diesel_prime.json](https://github.com/NREL/REopt_API/files/13756928/diesel_prime.json)
[pv_bess_primegen.json](https://github.com/NREL/REopt_API/files/13756929/pv_bess_primegen.json)
[web_tool_sample_site.json](https://github.com/NREL/REopt_API/files/13756930/web_tool_sample_site.json)
- It doesn't look like there's any accounting for Third-Party ownership? Is that just TBD? 
- In the old proforma (in _most but not all places_) fractions/percentages are reported as the whole number value (e.g. 8.3 instead of 0.083), which matters mostly because some of these fractions are then used in equations (e.g., the IRR equation). Since your values are all fractions, the equations are wrong. I think outputting them as fractions, modifying the formatting where needed, and changing the equations to not divide by 100 is probably the best path forward. 
- It does seem like the old code formatted things as "Numbers" or "Percentages" whereas this one all appears to have the "General" number format, or "Scientific" in some cases.

rathod-b:
@adfarth @Bill-Becker the branch now has proforma code integrated into it (i need to clean up the code a bit more). You can run a scenario, then supply its run_uuid to the new v3 proforma endpoint (GET), which returns an excel file of proforma in response.

Bill-Becker:
> Summary of some initial testing:
> 
> * 3/3 testing results jsons (attached below) produced errors in the code
>   [diesel_prime.json](https://github.com/NREL/REopt_API/files/13756928/diesel_prime.json)
>   [pv_bess_primegen.json](https://github.com/NREL/REopt_API/files/13756929/pv_bess_primegen.json)
>   [web_tool_sample_site.json](https://github.com/NREL/REopt_API/files/13756930/web_tool_sample_site.json)
> * It doesn't look like there's any accounting for Third-Party ownership? Is that just TBD?
> * In the old proforma (in _most but not all places_) fractions/percentages are reported as the whole number value (e.g. 8.3 instead of 0.083), which matters mostly because some of these fractions are then used in equations (e.g., the IRR equation). Since your values are all fractions, the equations are wrong. I think outputting them as fractions, modifying the formatting where needed, and changing the equations to not divide by 100 is probably the best path forward.
> * It does seem like the old code formatted things as "Numbers" or "Percentages" whereas this one all appears to have the "General" number format, or "Scientific" in some cases.

@rathod-b looks like the fractions vs percent is still an issue, where the values are fractions but they are being handled as percents in the equations which divide them by 100.

Also, and I mostly expected this, but I ran a CHP case, and the boiler and CHP fuel costs are not being handled (zeros). I think this is an update that needs to come from the REopt.jl proforma script(s).

rathod-b:
@Bill-Becker @adfarth Tested against 21 individual JSON files, most with success. TODO:
- Some files had marginal errors in NPV calculations but cash flows matched between JSON values and proforma values. This needs to be checked further.
- GHP tech outputs from JSON need to be reconciled with calculated proforma values.
- To make the proforma useful for Julia users, we should we create a new endpoint and function that generates an API inputs object given a REopt.jl inputs dictionary.
- Clean up the styles function.

PR #559, opened 2024-01-10
Fix summary endpt

Bill-Becker:
- Fixes an error from the /summary endpoint for off-grid evaluations when there is no ElectricTariffInputs
- Updates REopt.jl for the HiGHS solver environment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ Classify the change according to the following categories:
 ### Minor Updates
 #### Added 
 - Added the following BAU outputs:  lifecycle_chp_standby_cost_after_tax, lifecycle_elecbill_after_tax, lifecycle_production_incentive_after_tax, lifecycle_outage_cost, lifecycle_MG_upgrade_and_fuel_cost
+### Changed
+- Updated REopt.jl version to 0.39.1 along with updates to other dependencies
 #### Fixed
 - Fixed setting of default Generator `installed_cost_per_kw` so that user inputs are not overridden 
 - Avoid /summary endpoint error with off-grid runs where there is no ElectricTariff
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,14 +27,14 @@ Classify the change according to the following categories:
     ### Patches
 
 
-## Develop - 2024-01-07
+## v3.3.0
 ### Minor Updates
-#### Fixed
-- Fixed setting of default Generator `installed_cost_per_kw` so that user inputs are not overridden 
-
-## Develop - 2023-11-27
 #### Added 
 - Added the following BAU outputs:  lifecycle_chp_standby_cost_after_tax, lifecycle_elecbill_after_tax, lifecycle_production_incentive_after_tax, lifecycle_outage_cost, lifecycle_MG_upgrade_and_fuel_cost
+#### Fixed
+- Fixed setting of default Generator `installed_cost_per_kw` so that user inputs are not overridden 
+- Avoid /summary endpoint error with off-grid runs where there is no ElectricTariff
+
 ## v3.2.3
 ### Minor Updates
 ##### Changed
--- a/reoptjl/views.py
+++ b/reoptjl/views.py
@@ -856,7 +856,7 @@ def queryset_for_summary(api_metas,summary_dict:dict):
     if len(tariffOuts) > 0:
         for m in tariffOuts:
             if len(tariffInputs) > 0:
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -27,14 +27,14 @@ Classify the change according to the following categories:
...
 
 
-## Develop - 2024-01-07
+## v3.3.0
 ### Minor Updates
+#### Added 
+- Added the following BAU outputs:  lifecycle_chp_standby_cost_after_tax, lifecycle_elecbill_after_tax, lifecycle_production_incentive_after_tax, lifecycle_outage_cost, lifecycle_MG_upgrade_and_fuel_cost
 #### Fixed
 - Fixed setting of default Generator `installed_cost_per_kw` so that user inputs are not overridden 
+- Avoid /summary endpoint error with off-grid runs where there is no ElectricTariff
```
adfarth, line 36: Should we add the REopt.jl version update to the Changelog?

reoptjl/views.py
```diff
@@ -855,7 +855,10 @@ def queryset_for_summary(api_metas,summary_dict:dict):
     )
     if len(tariffOuts) > 0:
         for m in tariffOuts:
-            summary_dict[str(m.meta.run_uuid)]['year_one_savings_us_dollars'] = (m.year_one_energy_cost_before_tax_bau + m.year_one_demand_cost_before_tax_bau + m.year_one_fixed_cost_before_tax_bau + m.year_one_min_charge_adder_before_tax_bau) - (m.year_one_energy_cost_before_tax + m.year_one_demand_cost_before_tax + m.year_one_fixed_cost_before_tax + m.year_one_min_charge_adder_before_tax)
+            if len(tariffInputs) > 0:
+                summary_dict[str(m.meta.run_uuid)]['year_one_savings_us_dollars'] = (m.year_one_energy_cost_before_tax_bau + m.year_one_demand_cost_before_tax_bau + m.year_one_fixed_cost_before_tax_bau + m.year_one_min_charge_adder_before_tax_bau) - (m.year_one_energy_cost_before_tax + m.year_one_demand_cost_before_tax + m.year_one_fixed_cost_before_tax + m.year_one_min_charge_adder_before_tax)
```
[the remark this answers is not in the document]
Bill-Becker: Good points.  the "us_dollars" thing also seems like legacy v2 naming convention - but maybe we kept it instead of updating the web tool. If we changed it to `year_one_bill_before_tax` - `..._bau`, would that include the export benefit? Or do we still need to account for that separately?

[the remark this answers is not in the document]
adfarth: We would still need to account for that separately. The `year_one_bill_before_tax` does not account for the value of exports.

CHANGELOG.md
Bill-Becker: Yep, done, to v0.39.1

reoptjl/views.py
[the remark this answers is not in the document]
Bill-Becker: Updated to use `year_one_bill`

reoptjl/views.py
```diff
@@ -855,7 +855,10 @@ def queryset_for_summary(api_metas,summary_dict:dict):
     )
     if len(tariffOuts) > 0:
         for m in tariffOuts:
-            summary_dict[str(m.meta.run_uuid)]['year_one_savings_us_dollars'] = (m.year_one_energy_cost_before_tax_bau + m.year_one_demand_cost_before_tax_bau + m.year_one_fixed_cost_before_tax_bau + m.year_one_min_charge_adder_before_tax_bau) - (m.year_one_energy_cost_before_tax + m.year_one_demand_cost_before_tax + m.year_one_fixed_cost_before_tax + m.year_one_min_charge_adder_before_tax)
+            if len(tariffInputs) > 0:
+                summary_dict[str(m.meta.run_uuid)]['year_one_savings_us_dollars'] = (m.year_one_bill_before_tax_bau + m.year_one_export_benefit_before_tax_bau) - (m.year_one_bill_before_tax + m.year_one_export_benefit_before_tax)
```
adfarth, line 859: @Bill-Becker `m.year_one_export_benefit_before_tax` is actually a positive value if it's a benefit, so we should subtract this here if we want to account for it. However, I wonder if it makes more sense to just use `year_one_bill_before_tax` until we confirm how the value is used in the UI?

Bill-Becker: Oops, and I agree - I updated

adfarth (approved):
Approving and noting that the REopt version will get updated in the highs folder with later updates

PR #564, opened 2024-02-08
Update MACRS Bonus Depreciation to 60pct

adfarth:
- [x] CHANGELOG.md is updated

### Changed
- Changed **macrs_bonus_fraction** to from 0.80 to 0.60 (60%) for CHP, ElectricStorage, ColdThermalStorage, HotThermalStorage GHP, PV, Wind

Per Tax Cuts and Jobs Act of 2017: https://programs.dsireusa.org/system/program/detail/676 

Should align with https://github.com/NREL/REopt.jl/pull/350


reoptjl/models.py
```diff
@@ -5349,7 +5349,7 @@ class SIZE_CLASS_LIST(models.IntegerChoices):
     )
 
     macrs_bonus_fraction = models.FloatField(
-        default=1.0,
+        default=0.0,
```
Bill-Becker, line 5352: Really should be zero for generator?

Bill-Becker (approved):

reoptjl/models.py
adfarth: Doug has previously shared that the applicability of MACRS to generators is quite complicated: “depending on whether the generator is treated as attached to the building, whether it is eligible for a section 179 deduction, as well as whether it is >500kW. In practice, this makes for a range from 1 year to 40 years (!).” I think we could do something more sophisticated in the future, but should not assume 100% for now

Bill-Becker: Ok, I'm fine to change it to zero, or make it consistent with the other techs at 0.6. I'll let you decide.

adfarth: Doug and Dan have agreed with 0% so I feel good about that!

PR #565, opened 2024-02-13
Choose solver, add back V3 CI tests

Bill-Becker:
- Makes **one Julia environment** to avoid needing to update REopt.jl and other dependencies in multiple locations
- Makes **one http.jl** file with conditional loading for Xpress.jl, if it is installed
- Adds a **choice of solver** in `Settings.solver_name`, and useful error messages if trying to use Xpress without installation
- Changes all the GitHub Actions `test_job_endpoint.py` tests to use an open source solver so we can have full V3 **CI testing** again
- Updated the instructions in the licenseserver repo on github.nrel.gov
- Updated the Wiki in this repo with open source solver setup info
- Added instructions to REopt "All in One" document (NREL employees only)


.gitignore
```diff
@@ -117,7 +117,8 @@ docs/formulation/*log
 /.werf_secret_key
 
 # Customized solver setup and docker files
-julia_src/xpress/licenseserver
+julia_src/xpress/
```
adfarth, line 120: this directory should no longer exist, correct? Is this a check just in case someone hasn't deleted it?

julia_src/http.jl
```diff
@@ -1,82 +1,52 @@
...
-	else
-		@info "An error occured in the Julia code."
-		return HTTP.Response(500, JSON.json(error_response))
-	end
+include("os_solvers.jl")
+
+# Load Xpress only if it is installed, as indicated by ENV["XPRESS_INSTALLED"]="True"
+xpress_installed = get(ENV, "XPRESS_INSTALLED", "False")
+if xpress_installed == "True"
+    import Xpress  # Why not "using Xpress"?
```
adfarth, line 15: Why not? :)

reoptjl/test/test_job_endpoint.py
```diff
@@ -227,7 +227,7 @@ def test_superset_input_fields(self):
         r = json.loads(resp.content)
         results = r["outputs"]
 
-        self.assertAlmostEqual(results["Financial"]["npv"], 11323.01, places=0)
+        self.assertAlmostEqual(results["Financial"]["npv"], -258533.19, places=-3)
```
adfarth, line 230: Did just changing the solver change the result by this much?

adfarth:
@Bill-Becker 

- For the instructions in this Wiki, did you mean `-f` instead of `-d` in each of the new lines? (https://github.com/NREL/REopt_API/wiki/2.-Setup-for-Local-API-Development) 
- Since the xpress docker file also needs to be updated, I think we should recommend to the team to (1) delete the old `Dockerfile.xpress` and `licenseserver` directory and (2) follow the instructions on github.nrel.gov again (if you leave the old Dockerfile, you get an error due to the directory changing) 
- I tested locally with xpress and ran some of the tests and it worked with no issues

adfarth (approved):

.gitignore
Bill-Becker: Yes, it should no longer exist, but I'd rather keep this here as people transition over, just in case.

julia_src/http.jl
Bill-Becker: Updated to "using Xpress", and tested with V2 and V3

reoptjl/test/test_job_endpoint.py
Bill-Becker: Ah, no, it's because I had to fix all of the technology sizes to avoid a timeout with the open source solver. Fixing the sizes still tests the capabilities here, which also implies the NPV test is not needed.

.github/workflows/pull_request_tests.yml
```diff
@@ -27,14 +27,12 @@ jobs:
           NREL_DEV_API_KEY: ${{ secrets.NREL_DEV_API_KEY }}
```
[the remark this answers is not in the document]
Bill-Becker: Updated.

Bill-Becker:
> @Bill-Becker
> 
> * For the instructions in this Wiki, did you mean `-f` instead of `-d` in each of the new lines? (https://github.com/NREL/REopt_API/wiki/2.-Setup-for-Local-API-Development)
> * Since the xpress docker file also needs to be updated, I think we should recommend to the team to (1) delete the old `Dockerfile.xpress` and `licenseserver` directory and (2) follow the instructions on github.nrel.gov again (if you leave the old Dockerfile, you get an error due to the directory changing)
> * I tested locally with xpress and ran some of the tests and it worked with no issues
> * Just a few questions above but I am approving

Good catch and thanks for testing. I plan to go over this in the dev meeting tomorrow, and good to know we should just delete licenseserver and re-do those instructions.

PR #574, opened 2024-03-29
Use HiGHS solver by default and on production, include other OS solvers

Bill-Becker:
- Change the default solver to HiGHS instead of Xpress
- Remove Xpress for production API on NREL servers
- End-of-Life for v1 and v2 of the API which relied on Xpress on NREL servers
- Increase default timeout limit to 600 seconds (10 mins), and increase the max to 1200 seconds (20 mins) to allow slower solvers enough time to solve certain scenarios
- Increased the max optimality tolerance to 0.2 (20%) as a tool to debug long solve times and narrow in on a solution with decreased tolerance (keep default at 0.001 (0.1%)


reoptjl/models.py
```diff
@@ -233,10 +233,10 @@ class TIME_STEP_CHOICES(models.IntegerChoices):
         FOUR = 4
 
     timeout_seconds = models.IntegerField(
-        default=420,
+        default=600,
         validators=[
             MinValueValidator(1),
-            MaxValueValidator(420)
+            MaxValueValidator(1200)
```
zolanaj, line 239: I see no harm in going higher for API users, like 3600 - since harder problems might need more time.

zolanaj (approved):
Thanks @Bill-Becker!  One non-critical comment with the approval.

reoptjl/models.py
Bill-Becker: I'd be fine with a larger timeout for local/other-server running the API, but I don't want to clog up the NREL servers with super-long running runs. That's actually the main motivation for keeping this reasonably low.

zolanaj: Makes sense to me!  Thanks for clarifying.

PR #575, opened 2024-03-30
Use HiGHS solver by default, with Big M, and End-of-Life for v1 and v2

Bill-Becker:
- See PR https://github.com/NREL/REopt_API/pull/574


zolanaj (approved):
This builds locally and I was able to run test_job_endpoint just as a quick check.  Thanks @Bill-Becker!

CHANGELOG.md
```diff
@@ -26,6 +26,15 @@ Classify the change according to the following categories:
...
     ### Patches
 
+## v3.7.0
+### Minor Updates
+#### Changed
+- Default `Settings.solver_name` = `HiGHS`
+- See updates from REopt.jl v0.44.0: https://github.com/NREL/REopt.jl/releases/tag/v0.44.0
+- HiGHS, Cbc, and SCIP solvers use Big M notation constraints only in REopt.jl
+#### Deprecated
+- End-of-Life for v1 and v2 of the API for external/public interfacing from NREL servers
```
adfarth, line 36: @Bill-Becker should we include a bit more detail about what this means for users? E.g., Will users no longer be able to access results using v1 and v2 endpoints, or just will no longer be able to post to these endpoints?

reoptjl/models.py
```diff
@@ -296,7 +296,7 @@ class SOLVERS(models.TextChoices):
 
     solver_name = models.TextField(
         blank=True,
-        default=SOLVERS.XPRESS,
+        default=SOLVERS.HIGHS,
```
adfarth, line 299: This might not be critical for this PR, but if you use an empty string (below) for the solver_name input, it does not automatically select HiGHS and the error under "messages" is just `Unexpected Error...`

`post_1["Settings"]["solver_name"] = '' or ""`

adfarth (approved):
@Bill-Becker this look good to me! I just had two minor comments that could be worth addressing depending on priority. I spun up this branch locally and tested all of the solvers as well as the error messages from the v1 and v2 endpoints.

CHANGELOG.md
Bill-Becker: I made a discussion forum post about this, and added a link in both the API response error message and in the Changelog in https://github.com/NREL/REopt_API/pull/575/commits/089153d301053313bfcd01e030b1bd464a63cd67

reoptjl/models.py
Bill-Becker: Ok, yeah I'll punt on this for now. I don't think the user would enter an empty string accidentally in most cases.

Bill-Becker:
> @Bill-Becker this look good to me! I just had two minor comments that could be worth addressing depending on priority. I spun up this branch locally and tested all of the solvers as well as the error messages from the v1 and v2 endpoints.

Thanks Amanda! I addressed one of the two items, and punted on the other.

PR #591, opened 2024-07-08
Big M Constraint Updates, addition to /chp_defaults

Bill-Becker:
- Update Big M constraints to address bugs after indicator -> Big M conversions
  - https://github.com/NREL/REopt.jl/pull/401 
- Add `thermal_efficiency` as input to /chp_defaults endpoint to handle custom CHP prime movers which cannot normally supply steam
  - https://github.com/NREL/REopt.jl/pull/415

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,12 @@ Classify the change according to the following categories:
     ##### Removed
     ### Patches
 
+## Develop 2024-07-08
+#### Added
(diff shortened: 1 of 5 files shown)


zolanaj (approved):
Running the http endpoint, `thermal_efficiency` is accepted as input in REopt.jl and works as intended.

I made an update to the changelog, feel free to merge if you agree with the change.

PR #596, opened 2024-08-19
Custom-tables

bpulluta:
- [ ] CHANGELOG.md is updated

- Added a new functionality to generate a custom comparison table based on multiple `run_uuid` inputs.
- Introduced a new URL pattern to support multiple UUIDs in the URL, separated by semicolons.
- The `create_custom_comparison_table` view now processes multiple UUIDs and returns an Excel file as output.

### Other information:
- This change enhances the REopt API by allowing users to compare multiple scenarios in a single Excel output.
- The URL pattern for the comparison table has been updated to handle multiple UUIDs, ensuring backward compatibility with previous URL patterns.


Bill-Becker:
@bpulluta could you add an example way (or add functionality if it doesn't exist) to do the following (maybe in the "dummy" `other_custom_table`, which can be used as a sort of "template" to build a new custom table) :

1. Calculate a value for a row in the **BAU** column that is different from the way the other optimal columns are calculated? I.e. if we need to reference an `input` value for the BAU column, but an `output` for all the optimal scenarios
2. Do some math on multiple outputs (or inputs/outputs) and then use that value for the row
3. Hard code a value that we want to put in the cell, even if we know it's not coming from a REopt output or input
4. Format a selected row in a certain way, including either background/font or formatting the value like a % versus a float.

Again, amazing work to get this far of getting what I requested for the TASC. At this point, my requests are for expanding on the basic capability to make it extensible and easier to customize other tables which will be super helpful in the future too.

Bill-Becker (approved):

PR #601, opened 2024-09-24
Bug fix and cover a gap

rathod-b:
Fix a few bugs and cover a gap in re-linking portfolios and runs.

- [X] CHANGELOG.md is updated

Bug fix

Runs cannot be unlinked; once a run is unlinked from a portfolio, it cannot be re-linked to that portfolio.

Runs can be unlinked and successfully relinked to portfolios

No

### DIFF
--- a/reoptjl/views.py
+++ b/reoptjl/views.py
@@ -734,12 +734,23 @@ def link_run_uuids_to_portfolio_uuid(request):
                 for s in scenario:
                     s.portfolio_uuid = p_uuid
                     s.save()
+
+                # Existing portfolio runs could have been "unlinked" from portfolio
+                # so they are independent and show up in summary endpoint. If these runs
+                # are re-linked with a portfolio, their portfolio_id is updated above.
+                # BUT these runs could still show up under `Summary` results (niche case)
+                # because they are present in PortfolioUnlinkedRuns.
+                # Below, these runs are removed from PortfolioUnlinkedRuns
+                # so they are "linked" to a portfolio and do not show up under `Summary`
+                if PortfolioUnlinkedRuns.objects.filter(run_uuid=r_uuid).exists():
+                    obj = PortfolioUnlinkedRuns.objects.get(run_uuid=r_uuid)
+                    obj.delete()
             else:
                 # Stop processing on first bad run_uuid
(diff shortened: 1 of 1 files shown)


reoptjl/views.py
```diff
@@ -734,12 +734,23 @@ def link_run_uuids_to_portfolio_uuid(request):
                 for s in scenario:
                     s.portfolio_uuid = p_uuid
                     s.save()
+
```
rathod-b, line 737: @Bill-Becker Am I allowed to delete `PortfolioUnlinkedRuns` objects?

I realized that when we link a portfolio to an existing run, there can be a situation where the run is listed in `PortfolioUnlinkedRuns`, ie it was unlinked from a portfolio previously. In this case we do not want the run to show up in independent summary endpoint runs, which is why we'd have to delete that entry from `PortfolioUnlinkedRuns` table. I tested to see that an existing run can be deleted from `PortfolioUnlinkedRuns` and tied to a portfolio so it doesnt show up in independent summary runs, and then unlinked so it does show up.

reoptjl/views.py
```diff
@@ -810,6 +821,8 @@ def summary(request, user_uuid):
             run_uuid__in=[i.run_uuid for i in UserUnlinkedRuns.objects.filter(user_uuid=user_uuid)]
         ).only(
             'run_uuid',
+            'user_uuid',
```
rathod-b, line 824: Added these two fields in Summary response for informational and validation purposes. The UI can ignore these.

reoptjl/views.py
```diff
@@ -1287,7 +1302,7 @@ def unlink_from_portfolio(request, user_uuid, portfolio_uuid, run_uuid):
             elif runs[0].user_uuid != user_uuid:
                 return JsonResponse({"Error": "Run {} is not associated with user {}".format(run_uuid, user_uuid)}, status=400)
             else:
-                return JsonResponse({"Error": "Error in unlinking run {} from portfolio {}".format(run_uuid, portfolio_uuid)}, status=400)
+                pass
```
rathod-b, line 1305: Fixed a bug here so unlinking actually happens.

Bill-Becker (approved):

PR #612, opened 2024-11-04
Fix updating CHP.installed_cost_per_kw if scalar

Bill-Becker:
- Fix issue with CHP.installed_cost_per_kw not being an array when updating the inputs model object (which expects an array) in process_results.py, from Julia.
- For context, we rely on REopt.jl logic to assign CHP defaults, and then we have to update the API model fields after REopt.jl runs, in process_results.py. Since CHP.installed_cost_per_kw can be either a float or an array in REopt.jl, while in the API model field it's processed to always be an array, we need to check and convert to an array if it comes back from REopt.jl as a float.


rathod-b (approved):

rathod-b:
@Bill-Becker this looks good. We may need to merge develop into this branch and update the changelog.

PR #615, opened 2024-11-15
Hot Fix: ASHP defaults updating from REopt.jl

Bill-Becker:
PRs:
- https://github.com/NREL/REopt_API/pull/613


zolanaj (approved):
Thanks for this @Bill-Becker!  I really like the `prune_update_fields` method - that will allow for a lot more flexibility to add things in REopt.jl without causing issues for the API moving forward.

PR #604, opened 2024-09-30
Ghx defaults standardization

atpham88:
### Please check if the PR fulfills these requirements
- [ ] CHANGELOG.md is updated
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

### What kind of change does this PR introduce?
(Bug fix, feature, docs update, ...)

### What is the current behavior?
(You can also link to an open issue here)

### Does this PR introduce a breaking change?
(What changes might users need to make in their application due to this PR?)

PR #625, opened 2025-01-14
REopt.jl v0.50.0: Load Year Alignment

Bill-Becker:
See these updates from REopt.jl PRs:
- https://github.com/NREL/REopt.jl/pull/473

Also, add db migration file for previous PR merged in:
- https://github.com/NREL/REopt_API/pull/604


zolanaj (approved):
@Bill-Becker this looks good, thanks!  Reviewing for the ASHP force-dispatch update: handling of the new parameter looks correct - I like that `null=True, blank=False` so that the databse allows an empty value but the Web Tool does not - and the new defaults are working on my end.  I made a minor edit to the help text to match closer to what's on the REopt.jl side (tests passed when I sent to a separate branch and then I cherry-picked this over).  Under the premise that @adfarth approved the Ghx defaults update, this is good to merge.

PR #628, opened 2025-01-30
Updated Grid RE and Emissions

adfarth:
Inputs and outputs from: 
- https://github.com/NREL/REopt.jl/pull/426 

### Added 
- Add the following inputs to account for the clean or renewable energy fraction of grid-purchased electricity: 
  - ElectricUtility **cambium_cef_metric** to utilize clean energy data from NREL's Cambium database
  - ElectricUtility **renewable_energy_fraction_series** to supply a custom grid clean or renewable energy scalar or series
  - Site **include_grid_renewable_fraction_in_RE_constraints** - to allow user to choose whether to include grid RE in min max constraints
- Add the following outputs: 
  - ElectricUtility **annual_renewable_electricity_supplied_kwh**
  - Site **onsite_and_grid_renewable_electricity_fraction_of_elec_load**
  - Site **onsite_and_grid_renewable_energy_fraction_of_elec_and_thermal_load**
  - ElectricLoad **annual_electric_load_with_thermal_conversions_kwh** which calculates end-use electrical load after including electric consumption by heaters and chillers
- Add input option optimize_soc_init_fraction (defaults to false) to ElectricStorage, which makes the optimization choose the inital SOC (equal to final SOC) instead of using soc_init_fraction. The initial SOC is also constrained to equal the final SOC, which eliminates the "free energy" issue. We currently do not fix SOC when soc_init_fraction is used because this has caused infeasibility.
### Changed
- Change name of the following inputs: 
  - ElectricUtility input **cambium_metric_col** changed to **cambium_co2_metric**, to distinguish between the CO2 and clean energy fraction metrics
- Change name of the following outputs:
  - ElectricUtility **cambium_emissions_region** changed to **cambium_region**
  - Site **annual_renewable_electricity_kwh** changed to **annual_onsite_renewable_electricity_kwh**
  - Site **renewable_electricity_fraction** changed to **onsite_renewable_electricity_fraction_of_elec_load** 
  - Site **total_renewable_energy_fraction** changed to **onsite_renewable_energy_fraction_of_total_load**
- Change v3 endpoint "cambium_emissions_profile" to "cambium_profile"


julia_src/http.jl
```diff
@@ -597,7 +598,7 @@ HTTP.register!(ROUTER, "POST", "/erp", erp)
 HTTP.register!(ROUTER, "POST", "/ghpghx", ghpghx)
 HTTP.register!(ROUTER, "GET", "/chp_defaults", chp_defaults)
 HTTP.register!(ROUTER, "GET", "/avert_emissions_profile", avert_emissions_profile)
-HTTP.register!(ROUTER, "GET", "/cambium_emissions_profile", cambium_emissions_profile)
```
hdunham, line 600: What's the reason for changing this url?

reoptjl/models.py
```diff
@@ -437,32 +444,47 @@ class SiteOutputs(BaseModel, models.Model):
...
 
-    annual_renewable_electricity_kwh = models.FloatField(
+    annual_onsite_renewable_electricity_kwh = models.FloatField(
         null=True, blank=True,
         help_text=(
                   "Electricity consumption (incl. electric heating/cooling loads) that is derived from on-site renewable resource generation."
                   "Calculated as total annual RE electric generation, minus storage losses and curtailment, with the user selecting whether exported renewable generation is included). "
                   )
     )
-    renewable_electricity_fraction = models.FloatField(
```
hdunham, line 447: did you want to remove that from the help text before merging?

reoptjl/models.py
```diff
@@ -527,32 +549,46 @@ class SiteOutputs(BaseModel, models.Model):
...
                   "plus total annual thermal energy content of steam/hot water generated from renewable fuels (non-electrified heat loads)."
                   "The thermal energy content is calculated as total energy content of steam/hot water generation from renewable fuels,"
                   "minus waste heat generated by renewable fuels, minus any applicable hot water thermal energy storage efficiency losses."
                   "The denominator is calculated as total annual electricity consumption plus total annual thermal steam/hot water load."
                   )
     )
+    onsite_and_grid_renewable_electricity_fraction_of_elec_load_bau = models.FloatField(
+        null=True, blank=True,
+        help_text=(
+                  "Calculation is the same as onsite_renewable_electricity_fraction_of_elec_load, but additionally includes the renewable energy"
```
hdunham, line 581: maybe these help texts should reference the bau versions instead, i.e. onsite_renewable_electricity_fraction_of_elec_load_bau here and onsite_renewable_energy_fraction_of_total_load_bau below

reoptjl/test/posts/all_inputs_test.json
```diff
@@ -145,7 +145,8 @@
         "emissions_factor_CO2_decrease_fraction": 0.01174,
         "emissions_factor_NOx_decrease_fraction": 0.01174,
         "emissions_factor_SO2_decrease_fraction": 0.01174,
-        "emissions_factor_PM25_decrease_fraction": 0.01174
+        "emissions_factor_PM25_decrease_fraction": 0.01174,
```
hdunham, line 148: add other new inputs (renewable_energy_fraction_series and include_grid_renewable_fraction_in_RE_constraints) to this post too?

reoptjl/test/posts/central_plant_ghp.json
```diff
@@ -53,7 +53,12 @@
             "tess_ghx_minimum_timesteps_per_hour": 1,
             "max_sizing_iterations": 10,
             "init_sizing_factor_ft_per_peak_ton": 300.0,
-            "heat_pump_configuration": "WWHP"   
+            "heat_pump_configuration": "WWHP",
```
hdunham, line 56: what does this ghp stuff have to do with this PR?

reoptjl/models.py
```diff
@@ -1740,22 +1776,22 @@ class ElectricUtilityInputs(BaseModel, models.Model):
...
         blank=True,
         default = "lrmer_co2e",
         help_text=("Emissions metric used. Default is Long-run marginal emissions rate for CO2-equivalant, combined combustion and pre-combustion emissions rates. Options: See metric definitions and names in the Cambium documentation.")
     )
     cambium_start_year = models.IntegerField(
-        default=2024,
+        default=2025,
         validators=[
-            MinValueValidator(2023),
+            MinValueValidator(2025),
```
hdunham, line 1794: Should we really not allow users to give a past year for this? What if we want to model a system that already exists to validate or something?

reoptjl/test/posts/all_inputs_test.json
adfarth: added in [fefcfd4](https://github.com/NREL/REopt_API/pull/628/commits/fefcfd46935d02a16949db5cce286209604cc068)

reoptjl/test/posts/central_plant_ghp.json
adfarth: A month or so ago, An updated several defaults for GHP in [NREL/GhpGhx.jl](https://github.com/NREL/GhpGhx.jl/commit/fb3d8f5321ee8476952d81f1ef4d5d96ea20dc5e) and also the ghpghx/models.py file here. However, tests using this post continued to pass because the GhpGhx pkg in REopt.jl had not been updated until I recently ran "update" in the feature branch this is pointing to. So, the tests using this post were failing on this branch until I fixed the defaults to the old values

reoptjl/models.py
adfarth: I might be missing something, but users **can** give a year for this, it just has to be between 2025-2050 because those are the years for which Cambium has data. If they want to run a past year, they would need to supply their own custom emissions rates.

reoptjl/models.py
adfarth: yes, done in [ecd9e7d](https://github.com/NREL/REopt_API/pull/628/commits/ecd9e7d5467a9fc5d50f1cd050b7d953887de608)

reoptjl/models.py
adfarth: done in [ecd9e7d](https://github.com/NREL/REopt_API/pull/628/commits/ecd9e7d5467a9fc5d50f1cd050b7d953887de608)

julia_src/http.jl
adfarth: the reason is that you could now also use this to get the grid clean energy fraction profile, not just the emissions profiles. Do you think it's better to leave it, even if the name wouldn't be 100% accurate?

hdunham: no no fine to change it then

reoptjl/models.py
hdunham: I guess I was thinking that if cambium has time steps every 5 years from 2025 to 2050 and the nominal year also covers the 4 years preceding, that users would be allowed to provide a start year from 2021 to 2050. Maybe that's wrong though.

hdunham (approved):

PR #720, opened 2026-06-04
PostgreSQL v18 updates; Database restores

GUI:
This builds upon the deployment updates in https://github.com/NatLabRockies/REopt_API/pull/714, but with 2 main additions:

1. Updates to use the new PostgreSQL v18 database server, and also use PostgreSQL v18 in the Docker development environment.
2. Adds support for a database restore command that can be used for local development or for staging branches. This will restore a snapshot of production data to these environments, making it easier to test staging branches or perform local development with some real data in place.
    - Since the production data is large, this subset of data is limited to the past 15 days of all runs and then every 26 days from the past year (roughly equivalent to another 14 days of data, but spread across the year for more sampling). We could tweak these, but I was just trying to find a compromise of some subset that wasn't too large.
    - For local development, authorized NLR users can restore this snapshot by running `docker compose run --rm db-syncer rake db:data:restore`
    - For staging deployments, if you add the `deploy-db-restore` label to the deployment PR, then on the *next* deployment, it will restore with the latest snapshot once (the label will then be removed, but you can add it again to trigger another restore at a later date).
    - This will hopefully simplify staging testing and eliminate the need for some of the custom databases and deployment environments, since you can now test against real production data on any branch at any time. But still happy to chat more if there were other use-cases you all had.
    - The database restore solution here relies on some Ruby gems our team uses in our other projects. So I know it's maybe a bit funky to shoehorn this into this Python/Julia app, but it's fairly separate, so hopefully this won't matter a ton. It mostly boils down to just raw SQL, pg_dump, and pg_restore commands, this was just the easiest way to leverage the process our team uses in our Ruby projects. But I'm happy to revisit if this feels too odd or we run into any issues. But I'm hoping with the Docker abstraction, it's sort of irrelevant how it's written (you can just run that `docker compose run --rm db-syncer` locally and not worry about Ruby or anything else).


Bill-Becker:
@GUI when trying to do "docker compose run --rm db-syncer rake db:data:restore" locally, I get SSL certificate problem: self-signed certificate in certificate chain.

Bill-Becker:
Thoughts on why this deploy failed?
https://github.nrel.gov/TADA/REopt_API-deploy-mirror/actions/runs/495070

Was this deployed with deploy-db-restore label? I'm not sure how I can tell when it's removed after. I was trying to GET a run_uuid result from the production db that happened within the last 15 days through the staging pg-v18 branch, but it said the run_uuid was not in the db.

I do know we had a migration added after the May 15th deploy where you had to manually fix some migrations for the db snapshot, so we will need to do what you did before again.

Bill-Becker:
I just wanted to confirm that the celery (+julia) and julia pods are restarting nightly still. It doesn't look like they are based on the age of the pods there. Where if I compare to the master branch on staging, the pods seem to have restarted at ~1am as expected.

Bill-Becker (commented):
This looks good, but I left some comments separately/outside of this review to follow up on.

GUI:
> @GUI when trying to do "docker compose run --rm db-syncer rake db:data:restore" locally, I get SSL certificate problem: self-signed certificate in certificate chain.

Ah, I forgot to document this, but you'll need a `docker-compose.override.yml` file (which is gitignored) with some internal URL references. I'll followup via e-mail with that.

> Thoughts on why this deploy failed? https://github.nrel.gov/TADA/REopt_API-deploy-mirror/actions/runs/495070
> 
> Was this deployed with deploy-db-restore label? I'm not sure how I can tell when it's removed after. I was trying to GET a run_uuid result from the production db that happened within the last 15 days through the staging pg-v18 branch, but it said the run_uuid was not in the db.
> 
> I do know we had a migration added after the May 15th deploy where you had to manually fix some migrations for the db snapshot, so we will need to do what you did before again.

🤦‍♂️ My bad... I made a final commit to fix the CI test suite, but then I missed that inadvertently broke the deploys. This should be fixed by https://github.com/NatLabRockies/REopt_API/pull/720/commits/513c1141bb9b227360a7695fe0e7d7b8d705c450, though, so deploys should be running again.

Regarding the `deploy-db-restore` label and how that works, if the label is still on the PR, then the restore or deploy has *not* occurred yet. The label will be removed once the restore has successfully occurred. Here's a screenshot showing how I added the label this morning, and then once the restore successfully occurred, it was removed, and a reference to the latest deployment was added:
 
> Screenshot 2026-07-08 at 9 15 38 AM

But happy to chat more about any of this if there are still any questions.

> I just wanted to confirm that the celery (+julia) and julia pods are restarting nightly still. It doesn't look like they are based on the age of the pods there. Where if I compare to the master branch on staging, the pods seem to have restarted at ~1am as expected.

I believe only `master` pods get restarted nightly, it that correct? If so, then the new workflow at https://github.com/NatLabRockies/REopt_API/blob/pg-v18/.github/workflows/restart-celery-julia.yml should theoretically take over that job once all of this lands in `master`. But that's why these branched deployments aren't getting restarted yet. But if the old setup did restart branches, let me know, and I can poke around more.

Bill-Becker (approved):
We've worked though everything and I understand the updated process. Thanks!

